PR #3, opened 2021-07-01
create indexes to ADRs by tag

ripienaar:
To improve discoverability the list of ADRS
is rendered in the main repo README, one index
per tag.

Each ADR has some new metadata that defines tags
and a few other things, these are parsed by index.go
to generate the README.

A single ADR directory for all ADRs is again created


ripienaar:
Please see https://github.com/nats-io/nats-architecture-and-design/blob/b6f9f0991927019f1b718849baa9d21e6a75ebad/README.md for the final outcome.

ripienaar:
This PR is a response to @scottf stating:

```
Please just go to the repo and set it like you want and set the naming convention / standard. Present an issue asking for confirmation. We could be done with this already.
```

So here it is, please can those on the review list chime in.

We will then remove the ADRs from the server, I'll reconcile against the server with whatever ADRs @aricart has in flight atm.

aricart (commented):
LGTM

scottf:
It it possible to have the go code in a sub directory to prevent any confusion that this may be a code repo

ripienaar:
> It it possible to have the go code in a sub directory to prevent any confusion that this may be a code repo

It's a bit of a pain, medium term we should put this content on a https://gohugo.io/ website, this will have full text search and more. Then this repo is the source code for the site - end users see the site.

Changing this stuff into a sub dir is a bit annoying with how go "works" - you'd still end up with things in top of the repo.

scottf (approved):
LGTM

ripienaar:
I have added an issue template for new issues that will look like this one, we can iterate on this. 

https://github.com/nats-io/nats-architecture-and-design/issues/4

kozlovic (approved):
LGTM

.readme.templ
```diff
@@ -0,0 +1,42 @@
+![NATS](large-logo.png)
+
+# NATS Architecture And Design
+
+This repo is used to capture architectural and design decisions as a reference of the server implementation and expected client behavior.
+
+# Architecture Decision Records
```
matthiashanel, line 7: Suggestion: add a table listing all ADR not in a final ("Implemented", "Rejected") state. Below some seem to still be in Approved state, but are done IIRC.

matthiashanel (approved):
LGTM

.readme.templ
ripienaar: ok, will see if I can add it in a followup - keen to get merged so others can add issues and ADRs can come here.

We do need to be better at updating ADRs to Implemeneted.

wallyqs (approved):
LGTM, nice!

derekcollison (commented):
LGTM

PR #16, opened 2021-07-15
[modifying] ADR-1: adding unsigned 64 bit values

scottf:


ripienaar:
Seems to me this information would be better kept in the schemas where its more likely to be maintained and the first place someone looks.

We can do it like this:

```
diff --git a/schema_source/jetstream/api/v1/definitions.json b/schema_source/jetstream/api/v1/definitions.json
index 96570a7..351267c 100644
--- a/schema_source/jetstream/api/v1/definitions.json
+++ b/schema_source/jetstream/api/v1/definitions.json
@@ -5,6 +5,11 @@
   "description": "Shared definitions for the JetStream API",
   "type": "object",
   "definitions": {
+    "uint64": {
+      "$comment": "unsigned int 64",
+      "minimum": 0,
+      "maximum": 9007199254740991
+    },
     "stream_source": {
       "type": "object",
       "description": "Defines a source where streams should be replicated from",
@@ -16,8 +21,7 @@
         },
         "opt_start_seq": {
           "description": "Sequence to start replicating from",
-          "type": "integer",
-          "minimum": 0
+          "$ref": "#/definitions/uint64"
         },
         "opt_start_time": {
           "description": "Time stamp to start replicating from",
```

Then after building the schemas it looks like this:

```
            "opt_start_seq": {
              "description": "Sequence to start replicating from",
              "$comment": "unsigned int 64",
              "minimum": 0,
              "maximum": 9007199254740991
            },
```

We have many different ints in the api, int, int32, int64, uin64 etc, maintaining a document like this will be too hard - adding a 3rd place to put this information - improving the schemas is best.

ripienaar:
Above is the biggest number javascript will take, go wil encode up to `{"x": 18446744073709551615}` so for javascript a special decoder might be needed

ripienaar:
Things are a bit funky around top of uint64 so probably some languages will behave strangely - even here see I put `18446744073709551615` but JS turns that into `18446744073709552000`

scottf:
> Things are a bit funky around top of uint64 so probably some languages will behave strangely - even here see I put `18446744073709551615` but JS turns that into `18,446,744,073,709,552,000`

The only option would be to remove uint64 from the server. As you have mentioned JS only handles 53 bits of information, but the server is capable of sending a larger number, even as unlikely as it may be to have 18 quintillion (I looked that up!) messages.

I vote for both the schema and this document to be kept up to date, as this will be a quick reference for client developers.

ripienaar:
@derekcollison I think the vote here is any time you edit a NATS Server data type you should update this ADR. Please comment.

My proposal is as I maintain the Schemas I maintain a single point of truth after and any interested developer - and the NATS CLI - gains immediate benefit by having more correct information and validation in one place.

Clearly the NATS CLI is not going to be reading this ADR to validate user input - it can read schemas though, hence my vote is to improve the schemas.

ColinSullivan1:
I agree, the schema seems like the right place for field definitions like this.  Behavior around handling these messages would fit in an ADR though...

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+The most up-to-date information can be found in the [schema](https://github.com/nats-io/jsm.go/tree/main/schemas/jetstream/api/v1)
+
+## Schema Object Fields
+
+Object and list of fields within the object that are unsigned 64 bit values.
+
+#### `consumer_configuration`
+ 
+* `opt_start_seq`
+
```
kozlovic, line 22: Missing: `RateLimit       uint64        `json:"rate_limit_bps,omitempty"` // Bits per sec`

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+#### `consumer_create_request`
```
kozlovic, line 23: Really this one is the stream name and consumer_config.

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+* `opt_start_seq`
+
+#### `consumer_create_response`
```
kozlovic, line 27: The response is a consumer info, so see below.

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+* `consumer_seq`
+* `stream_seq`
+
+#### `consumer_info_response`
+
+* `opt_start_seq`
```
kozlovic, line 34: Anything in consumer_config + consumer/stream which is a sequence pair for Delivered and AckFloor.
But missing here is `NumPending     uint64          `json:"num_pending"`

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+* `consumer_seq`
+* `stream_seq`
+
+#### `consumer_list_response`
+
+* `opt_start_seq`
```
kozlovic, line 40: This is a list of consumer_info objects, so would not repeat fields here.

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+* `consumer_seq`
+* `stream_seq`
+
+#### `stream_configuration`
+ 
+* `opt_start_seq`
```
kozlovic, line 46: Yes, as part of StreamSource's.

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### `stream_create_request`
```
kozlovic, line 48: This is a stream configuration object really

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+* `opt_start_seq`
+
+#### `stream_create_response`
```
kozlovic, line 52: This is a stream_info really.

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+* `opt_start_seq`
+
+#### `stream_info_response`
+
+* `opt_start_seq`
+* `first_seq`
+* `last_seq`
```
kozlovic, line 60: Missing:
```
Msgs       uint64          `json:"messages"`
Bytes      uint64          `json:"bytes"`
Deleted    []uint64        `json:"deleted,omitempty"`
```
and from `Lost *LostStreamData`:
```
type LostStreamData struct {
	Msgs  []uint64 `json:"msgs"`
	Bytes uint64   `json:"bytes"`
}
```

kozlovic (dismissed):

adr/ADR-11.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### `stream_purge_request`
+
+* `seq`
+* `keep`
+
+## Message Meta Data
+
+The message meta data carried in the reply_to field of a JetStream message contains meta data. 
+All numerical fields are unsigned 64 bit values.
```
aricart, line 79: The real issue which is not really addressed here (here we are duplicating all the specifications) - is that if we are using JSON (which we are), then the max sequence number cannot exceed `9007199254740991`. This is within the positive range of numbers in JSON, which are only `2^53 - 1` - this means that we should be thinking on how to address this "limitation" by having a generational ID for the sequence, and the number to compute an effective sequence number regardless of the programming language used to access these values.

As it is custom parsing in the case of JavaScript is only possible if the value were expressed as a string. In all cases, we shouldn't duplicate the schema documentation (that is what the schema is).

aricart (approved):

aricart (dismissed):
we should deal with the JSON limit rather than spend time to document something that doesn't work as expected everywhere.

adr/ADR-11.md
scottf: 1. JSON doesn't care, it just defines a number as digits. Maybe Javascript cares, or schema tools care. Regardless, the server is already coded to use uint64. Even if it was changed to use int64, 2^63 -1 is still larger than 2^53 -1, which means almost every piece of numerical data in the system is at risk according to this. But Go, Java and .NET are handling these values fine.

Do you have a solution? 

2. I understand and agree that we should not duplicate data and the schema is the source of truth. If you all would prefer I will move all this to the issue I created to make sure that client parity is achieved, but I did edit the ADR to reference the schema as the most up to date information, and it doesn't hurt to call this issue out, as demonstrated by the conversation we are having.

adr/ADR-11.md
scottf: I've given this some thought. My intention was to reference the schema object, not the data structure. It does seem useful to present the data in both ways.

adr/ADR-11.md
ripienaar: >I understand and agree that we should not duplicate data and the schema is the source of truth. If you all would prefer I will move all this to the issue I created to make sure that client parity is achieved, but I did edit the ADR to reference the schema as the most up to date information, and it doesn't hurt to call this issue out, as demonstrated by the conversation we are having.

I think the conversation, thrashing out design etc is valuable and the ADR should focus on calling out what is needed in a broad sense and perhaps present some trategies for handling the large numbers.  Calling out individual fields - all of them the same - doesn't seem that useful and ensure that this ADR will be out of date soon. We might also propose that the server cap sequences to a number lower than uint64 while still storing the data in uint64 if we find supporting the full 64bit space is too problematic.

If we focussed on the problem as a whole and pull in call outs to the various other data types that also have to be handled correctly - for example specific int32's, languages should not send bigger numbers than int32 - that would make this ADR useful.

Combined with updating the source material (schemas) it would leave us in a really good place and this ADR being valuable

aricart: > ) JSON doesn't care, it just defines a number as digits. Maybe Javascript cares, or schema tools care. Regardless, the server is already coded to use uint64. Even if it was changed to use int64, 2^63 -1 is still larger than 2^53 -1, which means almost every piece of numerical data in the system is at risk according to this. But Go, Java and .NET are handling these values fine.

I think the IETF would disagree with you. JSON is a data exchange format, if you are not following the spec, you are doing something that looks like JSON but it is not JSON. In fields like headers, the sequence is not JSON, but on the info documents it is. Unfortunately just because one language can represent numbers unsigned in 64-bit space doesn't mean that the documents produce will be properly handled or parsed.

https://datatracker.ietf.org/doc/html/rfc7159#section-6

scottf: It does not require that. It just says its interoperable:
>Note that when such software is used, numbers that are integers and
   are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
   sense that implementations will agree exactly on their numeric
   values.
   
Which if we care about then the server must change. So again, the question is how should the server behave?

2^53 -1 (9,007,199,254,740,991) seems a sufficiently large number for a sequence (and will fit in a int64, making it easy for clients,) but there will have to be server code to manage what happens when the value gets to ...992

scottf:
@aricart @ripienaar I simplified the ADR based on our discussion. Do we need to do anything else?

ripienaar:
> @aricart @ripienaar I simplified the ADR based on our discussion. Do we need to do anything else?

There is no very little of value here in my oppionion, as in, nothing here really is actionable? I wonder if this could be an update to ADR-1 now instead? (ADR-1 needs a general revisit imo)

Who will update the schemas?

ripienaar:
As in I imagine as a entry point for new client developers ADR-1 should be the starting point, so makes sense to call this out as a general point there of gotchas to keep in mind?

scottf:
> I wonder if this could be an update to ADR-1

Sounds reasonable to me. I'll drop this branch/PR and combine my comments into that.

adr/ADR-1.md
```diff
@@ -97,6 +97,28 @@ Today the list API's just return `["ORDERS"]`, these will become:
 
 With the same `error` treatment when some error happens.
 
+### Numerical Values
+
+Some fields commuincated in the JetStream API are unsigned 64 bit values (uint64). These will be communicated as numbers in json. Clients must handle them, even if it means custom parsing. Commonly fields that represent a message sequence or id, or a field that holds a duration of nanoseconds are uint64 values. The most up-to-date information on specific fields can be found in the [schema](https://github.com/nats-io/jsm.go/tree/main/schemas/jetstream/api/v1).
```
ripienaar, line 102: lets also call out some are specifically 32bit and clients should be careful not to overflow them

ripienaar:
I think I've basically superceded this PR in https://github.com/nats-io/nats-architecture-and-design/pull/28, sorry I am not trying to dismiss your work - its just ADR-1 was frankly embarrasingly bad and of no use and out of date.  So adding to it in its current state would not really move the needle much.  So I rewrote it entirely.

scottf:
incorporated into https://github.com/nats-io/nats-architecture-and-design/pull/28

PR #18, opened 2021-07-19
new status unapproved

scottf:


main.go
```diff
@@ -28,7 +28,7 @@ type ADR struct {
 }
 
 var (
-	validStatus = []string{"Approved", "Partially Implemented", "Implemented", "Rejected"}
+	validStatus = []string{"Approved", "Partially Implemented", "Implemented", "Rejected", "Unapproved"}
```
ripienaar, line 31: What does this tatus mean?

scottf:
Interim status when an ADR is under discussion. It's not Approved yet. I don't know what Partially / Implemented are but they don't seem to apply. Rejected does not apply.

main.go
scottf: Interim status when an ADR is under discussion. It's not Approved yet. I don't know what Partially / Implemented are but they don't seem to apply. Rejected does not apply.

ripienaar: You can set it to approve in the PR - because while the PR is not merged "unapproved" is implied, once the PR is merged it's by default "approved" at the very least or one of the others if already in progress.

So a unmerged PR is Unapproved by default, no mered PR will be unapproved.  The status would thus never be used.

ripienaar: Essentally your PR proposes that a new ADR gets accepted with a certain status - Approved at least.

So the status in a PR should be what you propose to be on merged, you'd never propose unapproved?

ripienaar: We could probably remove Rejected, think it was a left over from what the initial ADR blog suggested.  `Superceded` would make sense as a status for when a later ADR replaces a earlier one.

Partially Implemented - that's essentially int64, several languages do it right already, others dont.
Implemented - whoever closes a ticket in this repo once all the languages support something should set the related ADR to that, it's done.

PR #20, opened 2021-07-20
[adding] adr to generate user jwt inside the client using scoped keys

matthiashanel:
Signed-off-by: Matthias Hanel <mh@synadia.com>


adr/ADR-12.md
```diff
@@ -0,0 +1,132 @@
...
+  "type": "user",
+  "version": 2
+ },
+ // public nkey portion of signingKey  
+ "sub": "UD44C3VDAEYG527W3VPY353B3C6LIWJNW77GJED7MM5WIPGRUEVPHRZ5"
+}
+```
+
+The body is first generated with `jti` set to `""`. 
+Then the `jti` value is computed as base32 encoded sha256 hash of the body.
```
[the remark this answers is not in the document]
aricart: A couple of notes on the base64 encoding:
- Padding characters (`=`) are removed
- `+` is changed to `-` and `/` is changed to `_`

Decoding obviously replaces `_` for `/` and `-` to `+`

adr/ADR-12.md
```diff
@@ -0,0 +1,132 @@
...
+To add a user with values similar to the earlier json body and sign it with the scoped signing key and output the json, execute:
+```bash
+> nsc add user --account ACC --name USER_NAME --tag PROVIDED_TAG1 --tag PROVIDED_TAG2  --expiry 2h --private-key anyrolename
+> nsc describe user --account ACC --name USER_NAME --json
+```
+
+An invocation of the utility function, generating a similar user would look like this:
+```go
+IssueUserJWT(accSigningKey, "ADECCNBUEBWZ727OMBFSN7OMK2FPYRM52TJS25TFQWYS76NPOJBN3KU4", "UD44C3VDAEYG527W3VPY353B3C6LIWJNW77GJED7MM5WIPGRUEVPHRZ5", "USER_NAME", 2*time.Hour, []string{"PROVIDED_TAG1", "PROVIDED_TAG2})
+```
```
aricart, line 132: Dialing all of this into perspective.
Client libraries won't generate the JWTs, the JWTs will be generated in response to the client performing some sort of authentication, which then yields a JWT that they can use (or handed out).

Looking at that need means that really the only languages, in addition, to go that should be necessary are _JavaScript, Java, and  .net (because it was already required). Requiring this to be available in _c_ or python, etc is extreme.

aricart (changes requested):

adr/ADR-12.md
```diff
@@ -0,0 +1,132 @@
...
+ // public nkey portion of signingKey 
+ "iss": "AACYICOAQMQ72EHT35R7LV6VFWMIVWFKWFE5P2JJ2TT674EO7DJTUHMM",
+ // base32 encoded sha256 hash of this json with "jti":""
+ "jti": "FRQCL7TPAIL6KHKPPPJS2YOHTANKNHTPLTX7STGPTGYZVGTOH2LQ",
+ // name provided or value of publicUserKey
+ "name": "USER_NAME",
+ "nats": {
+  // value of accountId 
+  "issuer_account": "ADECCNBUEBWZ727OMBFSN7OMK2FPYRM52TJS25TFQWYS76NPOJBN3KU4",
+  "pub": {},
```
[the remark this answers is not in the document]
matthiashanel: that's what nsc spit out, will remove

adr/ADR-12.md
```diff
@@ -0,0 +1,132 @@
...
+Static header json:
+```go
+{"typ":"JWT","alg":"ed25519-nkey"}
+```
+
+Commented body (//) json output that is subsequently turned into a jwt (treat not commented values as static):
+```go
+{
+ // expiration (when specified) time in unix seconds, derived from provided duration.
+ "exp": 1632077055,
```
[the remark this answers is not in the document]
matthiashanel: I will add the recommendation aspect to the function signature comment

adr/ADR-12.md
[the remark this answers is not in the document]
matthiashanel: will add that description

adr/ADR-12.md
matthiashanel: Tried to express that in the notes at the beginning.

matthiashanel: go should probably have that function as well

aricart (approved):
LGTM

ripienaar:
Need `go run main.go > README.md`

adr/ADR-12.md
```diff
@@ -0,0 +1,133 @@
+# JWT library free jwt user generation
+
+|Metadata|Value|
+|--------|-----|
+|Date    |2021-07-20|
+|Author  | @mhanel, @aricart |
+|Status  |`Proposed`|
+|Tags    |client, jwt|
```
ripienaar, line 8: better to have somthing like `accounts` or `security` instead of JWT - these tags are turned into indexes and its worth casting a slightly wider net so we dont end up with many single entry indexes

adr/ADR-12.md
```diff
@@ -0,0 +1,133 @@
+# JWT library free jwt user generation
+
+|Metadata|Value|
+|--------|-----|
+|Date    |2021-07-20|
+|Author  | @mhanel, @aricart |
+|Status  |`Proposed`|
```
ripienaar, line 7: Has to be one of Approved, Partially Implemented, Implemented, Rejected. In this case `Approved` is probably what you are aiming for once merged

matthiashanel: I figured I'll switch to the correct status once everyone reviewed and regenerate the readme as well.

adr/ADR-12.md
matthiashanel: changed to security

adr/ADR-12.md
matthiashanel: Actually, I'll set to Approved now and switch to Rejected in case this goes nowhere.

ripienaar:
LGTM

kozlovic (approved):
LGTM

PR #21, opened 2021-07-21
[adding] ADR 13: Describe Pull Subscribe internals

wallyqs:
Adding ADR that explains how `PullSubscribe` internals to fetch messages.

Signed-off-by: Waldemar Quevedo <waldemar.quevedo@gmail.com>


adr/ADR-11.md
```diff
@@ -0,0 +1,271 @@
...
+is as the error condition:
+
+```go
+for {
+	msgs, err := sub.Fetch(5)
+	if err != nil {
+		// Any error such as:
+		// 
+		// - client timeout
+		// - request timeout
```
wallyqs, line 238:
```suggestion
		// - request timeout (408)
```

adr/ADR-11.md
```diff
@@ -0,0 +1,271 @@
...
+		msg.Ack()
+	}
+}
+```
+
+#### Pull(1)
+
+For the case of fetching a single message, it is possible to optimize
+things by making the first fetch request as a no wait request instead
+and by preparing a new style like request/response handler.
```
wallyqs, line 257:
```suggestion
and by preparing a new style like request/response handler using a wildcard
subscription.  This will result in less chatty protocol and also works better with
topologies where JetStream is running as a leafnode on the edge.
```

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+{"batch":5,"no_wait":true}
+HMSG _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 8  28 28
+NATS/1.0 404 No Messages
+UNSUB 1 4
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 32
+{"expires":4990000000,"batch":5}
+MSG hello 1 $JS.ACK.bar.dur.2.29034.29041.1626845015078897000.0 4
+ping
+# Only 1 out of 5 messages receives, so client goes away and unsubscribes.
+UNSUB 1
```
kozlovic, line 228: So does that mean that the library sets a subscription for a batch and if the batch is full, we know the server does not send more, but otherwise we "timeout" the subscription if not all messages in the batch were received?
Most libs don't have a wait to timeout an async subscription, so timeout is usually with sync subscription. If that is the case, then will have to keep track of time already spent dequeueing some messages from the sub (that is, timeout will decrease overtime).

And to be clear, in the case above where 1 message was received (out of 5), this is not considered and error and that single message is returned by Fetch(5), right?

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+SUB _INBOX.WvaJLnIXcj8Zf5SrxlHMTS  1
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 26
+{"batch":5,"no_wait":true}
+HMSG _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 8  28 28
+NATS/1.0 404 No Messages
+UNSUB 1 4
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 32
+{"expires":4990000000,"batch":5}
+MSG hello 1 $JS.ACK.bar.dur.2.29034.29041.1626845015078897000.0 4
+ping
```
kozlovic, line 226: Not sure what `ping` means here.

kozlovic (commented):
Have some comments and questions.

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+	for _, msg := range msgs {
+		msg.Ack()
+	}
+}
+```
+
+When implementing `PullSubscribe` there are two main cases to
+consider: `Fetch(n)` and `Fetch(1)`.
+
+#### Pull(n)
```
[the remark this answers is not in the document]
wallyqs: yes they are synonyms, meant to search replace everything to Pull but missed these.  `Fetch` is the name of the Go client API to get a collection of messages

adr/ADR-13.md
wallyqs: it is just the payload, will update to something else

kozlovic: got it, sorry.

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+`_INBOX.example`, when the message got delivered the subject was
+rewritten into `bar` which is the subject of the message that is
+persisted in JetStream.
+
+A pull request for the next message, will linger until there is no
+more interest in the subject, a client is disconnected or the batch size is filled.
+Each pull request will increase the `num_awaiting` counter for a consumer of
+inflight pull requests [2].  At most, a consumer can only have 512
+inflight pull requests, though this can be changed when creating the
+consumer with the `max_waiting` option [1]:
```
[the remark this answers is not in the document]
wallyqs: just informational, when creating the consumer one can increase the limit of 512 by modifying `max_waiting`.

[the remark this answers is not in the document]
kozlovic: ok, thanks

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+The result of a no wait pull request is a guaranteed instant response
+from the server that will be either the next message or an error,
+where an error could be a server error such as `503` in case JS
+service is not available. Most commonly, the result of a no wait pull
+request will be a `404` no messages error:
+
+```shell
+HMSG _INBOX.x7tkDPDLCOEknrfB4RH1V7.OgY4M7 2  28 28
+NATS/1.0 404 No Messages
+```
```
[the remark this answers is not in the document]
wallyqs: no, it should then fall back to do a pull request that waits and blocks until there are new messages or there is a timeout, so the same request but without `no_wait`

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+		continue
+	}
+	for _, msg := range msgs {
+		msg.Ack()
+	}
+}
+```
+
+When implementing `PullSubscribe` there are two main cases to
+consider: `Fetch(n)` and `Fetch(1)`.
```
wallyqs, line 168:
```suggestion
consider: `Pull(n)` and `Pull(1)`.
```

adr/ADR-13.md
```diff
@@ -0,0 +1,290 @@
...
+done to try to get the messages that may already be available as
+needed.  If there are no messages (a 404 status message error by the
+JetStream server), then a long pull request is made:
+
+```shell
+SUB _INBOX.NQ2uAIXd4GoozKOTfECtIg  1
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 26
+{"batch":5,"no_wait":true}
+
+# Result of no wait request if there are no messages
```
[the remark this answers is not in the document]
wallyqs: it would then wait for all the messages in the batch to be delivered or the request will timeout waiting for the complete batch of messages.

adr/ADR-13.md
wallyqs: In case of the Go client, yes only 1 message out of the 5 collected would be returned, but no `nats: timeout` error is returned for example

adr/ADR-13.md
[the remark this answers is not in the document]
kozlovic: So this is not really an "option" that the user set. This is done internally, always, no? The use of the options no_wait, etc.. got confusing because I thought these were options that the user can pass to the "Fetch" api, but this does not seem to be the case then.
If it is, then I don't understand that an option "no_wait" would lead to... a wait for messages ;-)

adr/ADR-13.md
wallyqs: a sync subscription might be fine since for batches larger than 1, a new subscription has to be made for each request, then unsubscribe once there is a timeout or the complete set of messages was delivered (just like old style requests just that instead of expecting single response expecting now many)

adr/ADR-13.md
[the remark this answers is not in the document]
kozlovic: Ok, so `no_wait` first says there are no messages, then we go to "long" wait and again batch is 5, but only 4 are avail, so library will wait up to the timeout and then will return no error since there is at least 1 message to return to the user.
(what if there is no timeout specified, do we wait for ever? in here we have 4 messages out of 5 so no message will ever be returned. So in other words, is a timeout mandatory for Fetch() calls?)

adr/ADR-13.md
kozlovic: I think I understand.

adr/ADR-13.md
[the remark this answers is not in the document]
wallyqs: That's right, there is no need for an exposed `no_wait` option since it is just used internally to 'start the session' of being delivered messages by the server. `No Wait` in this context means 'get any response from the server asap' which is either the batch of messages ready to be delivered or an error.

adr/ADR-13.md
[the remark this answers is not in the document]
wallyqs: By default in nats.go the timeout is 5 seconds, but users can pass a context without timeout and drive the cancellation on their own, etc... but thinking that maybe we should make the timeout mandatory.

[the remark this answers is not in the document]
wallyqs: Some other reasons that `no_wait` is used as well is to further help for the `Pull(1)` use case (which likely is the most common), 'no wait' requests do not become long pull requests, run into 408 Request Timeout errors since not accounted as inflight pull request since meant to be instant.
Also just found a mistake in the unsub, it should have been `+1`...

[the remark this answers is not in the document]
wallyqs: fixed now

kozlovic (approved):
LGTM

adr/ADR-13.md
```diff
@@ -0,0 +1,291 @@
...
+A pull based consumer is a type of consumer that does not have a
+`delivery subject`, that is the server _does not know_ where to send
+the messages.  Instead, the clients have to request for the messages
+to be delivered as needed from the server.  For example, given a stream `bar`
+with the consumer with a durable name `dur` (all pull subscribers have to be
+durable), a pull request would look like this in terms of the
+protocol:
+
+```shell
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.x7tkDPDLCOEknrfB4RH1V7.UBZe2D 0
```
ripienaar, line 27: I'd avoid showing this basic form, generally we should never make pulls that doesnt have `no_wait` or `expires` since they cause confusing behavior and pile ups on the server.  The design you outline does the right thing - but someone might see this example and think its a recommendation or something

ripienaar (approved):
LGTM except the minor comment

adr/ADR-13.md
wallyqs: sounds good, updated to an example with `expires` instead

adr/ADR-13.md
```diff
@@ -0,0 +1,291 @@
...
+SUB _INBOX.WvaJLnIXcj8Zf5SrxlHMTS  1
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 26
+{"batch":5,"no_wait":true}
+HMSG _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 8  28 28
+NATS/1.0 404 No Messages
+UNSUB 1 6
+PUB $JS.API.CONSUMER.MSG.NEXT.bar.dur _INBOX.WvaJLnIXcj8Zf5SrxlHMTS 32
+{"expires":4990000000,"batch":5}
+MSG hello 1 $JS.ACK.bar.dur.2.29034.29041.1626845015078897000.0 4
+ping
```
wallyqs, line 227:
```suggestion
helo
```

PR #22, opened 2021-07-21
Hostname resolution in client library

kozlovic:
Signed-off-by: Ivan Kozlovic <ivan@synadia.com>


ripienaar:
LGTM

ripienaar (approved):

adr/ADR-13.md
```diff
@@ -0,0 +1,38 @@
...
+## Prior Work
+
+The Go client is doing host name resolution as shown [here](https://github.com/nats-io/nats.go/blob/2b2bb8f326dfdd2814ba6d59c59b562354b1af30/nats.go#L1641)
+and then shuffle this list (unless the `NoRandomize` option is enabled) as shown [here](https://github.com/nats-io/nats.go/blob/2b2bb8f326dfdd2814ba6d59c59b562354b1af30/nats.go#L1663)
+
+## Design
+
+When the library is about to create a TCP connection, if given a host name (and not an IP), a name resolution must be performed.
+
+If the list has more than 1 IP returned, it should be randomized, unless the existing `NoRandomize` option is enabled.
```
scottf, line 23: consider that there are 2 host names each with multiple ip addrs. Also maybe there are other hosts / ips in the full server list. The entire list of ips should be resolved then randomized, which is congruent with the existing randomize behavior. I don't know if this needs to be explicitly called out. It may be useful to make this adr a combination of resolution and randomization. https://github.com/nats-io/nats.go/issues/565

kozlovic: I think that this ADR is really about: "for a given URL we are trying to create a TCP connection to", so this is in addition to any randomization that is done at a higher level regarding list of URLs the client can connect to.

Trying to get all IPs of all "server URLs" and working on that list may not be best in that with the current ADR, there is a chance for the DNS to return new IPs when the connection attempt is made. In your proposal, the resolution would be done and the library would be stuck with that.
It would also make it very difficult for the library to update that list of IPs when a server URL (from auto-discovery, async INFO) is removed from the list.

derekcollison (approved):
LGTM

ColinSullivan1 (approved):
LGTM.

adr/ADR-13.md
scottf: Ok I think I understand. We already randomized the list and we are now about to connect to a hostname (versus an ip). That hostname gives us more than 1 ip address, so randomly choose one of them. Can we get that into the doc?

kozlovic: @scottf I think I am already documenting that in here:
```
When the library is about to create a TCP connection, if given a host name (and not an IP), a name resolution must be performed.

If the list has more than 1 IP returned, it should be randomized, unless the existing `NoRandomize` option is enabled.
We could introduce a new option specific to this IP list as opposed to the server URLs provided by the user.

Then the connection should happen in the order of the shuffled list and stop as soon as one is successful.
```
Not sure what you want me to change? Is it the bit of the "randomized list (of server URLs)"? Otherwise feel free to submit PR to modify, since this has been merged.

PR #27, opened 2021-07-21
[adding] ADR-18: URL support for all client options

scottf:


adr/ADR-14.md
```diff
@@ -0,0 +1,28 @@
...
+
+Anything tuning connection behavior, which might be used as an option on establishing a connection, should be specifiable in a URL.  Anything which doesn't fit into authority information in the URL should probably be "query parameters", `?opt1=foo&opt2=bar` with the documentation establishing the option names and the behavior for unrecognized values for each option.  Unrecognized options should be ignored.  It's possible that some options should be `#fragopt1=foo&opt2=bar` instead and we should clearly define where we draw the line.  Eg, "if it's not sent to the server but is reconnect timing information, it should be `#fragment`" or "for consistency we use `?query` for them all".
+
+Things which should be configurable include, but are not limited to:
+
+ * OCSP checking status
+ * JetStream Domain
+ * Various timeouts
+ * TLS verification level (if we support anything other than verify always)
+ * Server TLS cert pinning per `hex(sha256(spki))`
```
matthiashanel, line 28: This in itself is a separate feature that does not exist in clients yet.
When we add it, it needs to come with a dedicated option as well.
We should probably support a list vs a single value.

adr/ADR-14.md
```diff
@@ -0,0 +1,28 @@
...
+Making explicit comma-separated lists of URLs, vs of hostnames within a URL, and ensuring that is compatible across all clients is included, plus order randomization.
+
+Anything tuning connection behavior, which might be used as an option on establishing a connection, should be specifiable in a URL.  Anything which doesn't fit into authority information in the URL should probably be "query parameters", `?opt1=foo&opt2=bar` with the documentation establishing the option names and the behavior for unrecognized values for each option.  Unrecognized options should be ignored.  It's possible that some options should be `#fragopt1=foo&opt2=bar` instead and we should clearly define where we draw the line.  Eg, "if it's not sent to the server but is reconnect timing information, it should be `#fragment`" or "for consistency we use `?query` for them all".
+
+Things which should be configurable include, but are not limited to:
+
+ * OCSP checking status
+ * JetStream Domain
+ * Various timeouts
+ * TLS verification level (if we support anything other than verify always)
```
matthiashanel, line 27: Not sure how deep we want to go here. 
We do not support that as direct nats option (in go at least)
But you can set a tls config. So perhaps need to check which values in there make sense to have. 

That said, I would not expose values that we wouldn't want users to use in production.
Deferring to @philpennock here, but I would NOT make `InsecureSkipVerify` settable via this.

matthiashanel (approved):
feature makes sense to me. LGTM

ripienaar:
The LInt check failed. Why was this merged on red?

PR #28, opened 2021-07-22
[modifying] ADR-1:  update jetstream api adr for reality

ripienaar:

### DIFF
--- a/adr/ADR-1.md
+++ b/adr/ADR-1.md
@@ -164,7 +164,7 @@ If you're writing a client and have JSON Schema validators to hand you can acces
 ## Data Types
 
 The JetStream API has many data types, generally the JSON Schema tries to point this out, though we have some gaps in the data for sure.
-This section explore a few of the types in details.
+This section explores a few of the types in details.
 
 ### Sequence related unsigned 64bit integers
 
@@ -201,6 +201,8 @@ Some fields are thus going to be dynamically capped to the server architecture, 
 }
 ```
 
+Note while this is an unsigned integer, schema documents may list a different minimum than the minimum for the type.
+
 ### 32bit Integers
 
 Similar to the 64bit integers, we just have a 32bit one, care should be taken not to overflow this number when sending data to the server, here's an example.
@@ -216,6 +218,8 @@ Similar to the 64bit integers, we just have a 32bit one, care should be taken no
 }
 ```
 
+Note while this is an unsigned integer, schema documents may list a different minimum than the minimum for the type. Here the minimum is `-1`.
+
 ### Time Durations
 
 Some fields, like the maximum age of a message, are expressed as durations like `1 day`.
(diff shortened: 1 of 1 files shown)


ripienaar:
An attempt to make this first ADR be more like a solid intro to developing against the API, previously it was essentially just useless.

Feedback welcome.

adr/ADR-1.md
```diff
@@ -4,152 +4,280 @@
...
-error handling should look at the `code` which will be HTTP like, 4xx human error, 5xx server error etc. Finally, the `error_code`
-indicates the specific reason for the 404 - here `10059` means the stream did not exist, helping developers identify the
-real underlying cause. 
+## Schemas
 
-More information about the `error_code` system can be found in [ADR-7](0007-error-codes.md).
+All requests and responses that are in JSON format have JSON Schema Draft 7 documents describing them.
 
-Ideally the error response includes a minimally valid body of what was requested but this can be very hard to implement correctly.
+For example the request to create a stream has the type `io.nats.jetstream.api.v1.stream_create_request` and the response
```
kozlovic, line 122: Should library written now look for `v1.` in the response and fail if not found? Meaning is that expected that if it one day becomes `v2.` it means that the client will not be able to understand the response?

kozlovic (approved):
LGTM, just have question about expected client behavior with version in response type.

adr/ADR-1.md
ripienaar: Some people believe we'll never see version 2 - we're only adding fields ever.  

Generally I check the full type matches exactly one that I have a known type for - though granted I've fully incorporated 100% validated requests and responses in the API and is probably more than is realistic.

In the CLI/Terraform/GitHub Actions any response that does not validate 100% of does not have a known schema would cause a panic and I also check requests to the server against the schemas.

The real answer is somewhere in the middle - I'd check the full type field personally at least.

kozlovic: So to make sure I understand, if I send a "stream_create_request", then when the library gets a response, it should fail if the response's type is not "io.nats.jetstream.api.v1.stream_create_response"?

ripienaar: I fail in that case yes, is it the 100% right thing to do? Not sure, I think its a very strong signal that things are in a bad way somewhere and someone should know.  So that's what the CLI does.

Clients might want to be a bit more relaxed about it, seems risky? But if you are getting a response today without that type I can guarantee you things are very broken.

scottf: It would be interesting to see what clients have checked the response type. Java and .NET have not. If this is important, its relatively trivial to do but it's work to be done. Should I add a line to the parity sheet and and open an issue?

ripienaar: For today I dont think it's a problem - we're way off from having a v2 - hopefully never.  I wouldnt loose sleep on it, we can punt till we get to v2.

adr/ADR-1.md
```diff
@@ -4,152 +4,280 @@
...
 
-While implementing this in JetStream the following pattern emerged:
+## Data Types
 
-```go
-type JSApiResponse struct {
-	Type  string    `json:"type"`
-	Error *ApiError `json:"error,omitempty"`
+The JetStream API has many data types, generally the JSON Schema tries to point this out, though we have some gaps in the data for sure.
+This section explore a few of the types in details.
```
ripienaar, line 167: I'll add a bit here - since the schemas will show different minimums for example many signed int64s have a minimum value of -1 for example in the schema, clients must do the right thing there.

adr/ADR-1.md
ripienaar: For the CLI I would via validation also detect if I am running against a server returning new fields and it would fail. 

I am not sure what clients can do as we have no negotiation around api versions etc. These fields are the only guide - and we have had some breaking changes already. 

I don’t know what the answer is for languages who can’t do schemas tbh :)

PR #30, opened 2021-07-23
Extend the KV model to support range watches

ripienaar:
This extends the kv model, allowing `.` in a name and
ranged watches matching a subset of data in a bucket.

### DIFF
--- a/adr/ADR-8.md
+++ b/adr/ADR-8.md
@@ -31,8 +31,8 @@ Initial feature list:
  * Encoders and Decoders that transforms both keys and values
  * A read-cache that builds an in-memory cache for fast reads
  * read-after-write safety unless read replicas used
- * Valid keys are `\A[-/_a-=zA-Z0-9]+\z` after encoding
- * Valid buckets are `^[a-zA-Z0-9_-]+$`
+ * Valid keys are `\A[\./-_a-zA-Z0-9]+\z` after encoding
+ * Valid buckets are `\A[a-zA-Z0-9_-]+\z`
  * Custom Stream Names and Stream ingest subjects to cater for different domains, mirrors and imports
  * Key starting with `_kv` is reserved for internal use
  * CLI tool to manage the system as part of `nats`, compatible with client implementations
@@ -95,7 +95,7 @@ type Status interface {
 	// BucketLocation returns a string indicating where the bucket is located, meaning dependent on backend
 	BucketLocation() string
 
-	// Keys returns a list of all keys in the bucket - not possible now
+	// Keys returns a list of all keys in the bucket - not possible now except in caches
 	Keys() ([]string, error)
 
 	// BackingStore is a backend specific name for the underlying storage - eg. stream name. Used so that ops tooling can know what to Backup for example
@@ -122,8 +122,9 @@ type RoKV interface {
 	// History retrieves historic values for a key
 	History(ctx context.Context, key string) ([]Entry, error)
 
-	// WatchBucket watches the entire bucket for changes, all keys and values will be traversed including all historic values
-	WatchBucket(ctx context.Context) (Watch, error)
+	// WatchBucket watches the entire bucket for changes, all keys and values will be traversed including all historic values.
+	// WatchOps can supply modifiers to perform range watches
+	WatchBucket(ctx context.Context, opts ...WatchOpts) (Watch, error)
 
 	// Watch a key for updates, the same Entry might be delivered more than once
 	Watch(ctx context.Context, key string) (Watch, error)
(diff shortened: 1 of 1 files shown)


adr/ADR-8.md
```diff
@@ -31,8 +31,8 @@ Initial feature list:
  * Encoders and Decoders that transforms both keys and values
  * A read-cache that builds an in-memory cache for fast reads
  * read-after-write safety unless read replicas used
- * Valid keys are `\A[-/_a-=zA-Z0-9]+\z` after encoding
- * Valid buckets are `^[a-zA-Z0-9_-]+$`
+ * Valid keys are `\A[\./-_a-zA-Z0-9]+\z` after encoding
```
derekcollison, line 34: What is the effect of this change? So I do not have to pull out my regex fu. ;)

adr/ADR-8.md
```diff
@@ -31,8 +31,8 @@ Initial feature list:
...
+ * Valid buckets are `\A[a-zA-Z0-9_-]+\z`
```
derekcollison, line 35: Same here.

adr/ADR-8.md
```diff
@@ -122,8 +122,9 @@ type RoKV interface {
 	// History retrieves historic values for a key
 	History(ctx context.Context, key string) ([]Entry, error)
 
-	// WatchBucket watches the entire bucket for changes, all keys and values will be traversed including all historic values
-	WatchBucket(ctx context.Context) (Watch, error)
+	// WatchBucket watches the entire bucket for changes, all keys and values will be traversed including all historic values.
```
derekcollison, line 125: So if I watch a key I get the old values first? Seems like maybe opt-in to that but default is last value.

adr/ADR-8.md
```diff
@@ -122,8 +122,9 @@ type RoKV interface {
...
+	// WatchOps can supply modifiers to perform range watches
+	WatchBucket(ctx context.Context, opts ...WatchOpts) (Watch, error)
 
 	// Watch a key for updates, the same Entry might be delivered more than once
 	Watch(ctx context.Context, key string) (Watch, error)
```
derekcollison, line 130: Instead of watch bucket above why not just allow wc on Watch for key?

adr/ADR-8.md
ripienaar: Allows `.`, though thinking about it I need to improve this a bit to avoid a leading `.`.

adr/ADR-8.md
ripienaar: just avoids some newline related issues when validating the regex - intent stays the same but now the regex correctly captures the intent :)

adr/ADR-8.md
ripienaar: yes, only if you watch the entire bucket.  If you did a specific key watch you only get the most recent value. We would need something like Subject Filter on `foo.*` with a starting position of every last message in every subject - awkward. 

I think the sweet spot for history isnt going to be like 100s of historic values, but 10 (I might cap it at 10), so should be fine.

adr/ADR-8.md
ripienaar: because a key watch behaves quite differently from a bucket watch - as above bucket watch gives you history. So a scoped bucket watch more describes what is possible with this subset, since JetStream has no feature to give latest per subject.

adr/ADR-8.md
derekcollison: Starting position of Last, but LastBySubject similar to what we added for msg limits?

adr/ADR-8.md
derekcollison: Do we need history though in most cases or do we need to consider adding in the config item from above?

adr/ADR-8.md
ripienaar: If this existed I wouldnt complain :)

adr/ADR-8.md
ripienaar: With above `LastBySubject` we could drop the history behavior by default but keep it as an opt-in should someone need it, would be quite nice without the history.  Would definitely not mind having that feature.

derekcollison: ok may do some R&D..

ripienaar: 1790e73779501cd0ea9b157333a74435e722814b updates the doc to remove the `WatchBucket()` function entirely and to support ranges on watch, so `Watch(ctx, "") // bucket` or `Watch(ctx, "x.>") // range` or usual `Watch(ctx, "x".y")`

Now both types of watch - bucket, key, range etc - all behave the same thanks to `last_per_subject`

aricart (approved):
LGTM - we discussed other usage patterns that we'll review offline from this PR.

PR #36, opened 2021-08-03
[adding] ADR-16 Consumer groups

ripienaar:


ripienaar:
@derekcollison this is a rough outline for partitioned clients or client groups - we can figure out the name too.

For now just capturing the problem statement, we should go from here and work on designing a solution.

adr/ADR-16.md
```diff
@@ -0,0 +1,135 @@
...
+|---|-------|-------|--------|
+|1  |sensor.1.temp|28.1|A|
+|2  |sensor.2.temp|20|B|
+|3  |sensor.1.humidity|60|A|
+|4  |sensor.2.humidity|55|B|
+|5  |sensor.3.temp|30|A|
+|6  |sensor.3.humidity|70|A|
+
+When doing summarization of these metrics 2 consumers could be created, and consuming messages in parallel, as long as 
+all the messages for a given sensor always goes to the same consumer. Each of the consumers could have a `max_ack_pending`
```
aricart, line 38: I think what you are describing is an extension of what the KV is already able to do (somewhat). If we had the ability to extract all the subjects in a stream _without_ iterating over the last message for each subject, you could have a controller process watching over keys and creating specific consumers to process a specific subject's history. This effectively allows you to implement a map/reduce - also if you are doing summations/averages, there's no need to have ordering. In that case the algo prob just can just use a KV to store sum and counts, and update on that.

ripienaar: > you could have a controller process watching over keys and creating specific consumers to process a specific subject's history

That would end up with a consumer per device wouldnt it? Hence the need for a server level help with this, we'd rather have 10 consumers and have the server funnel the traffic into the 10 consumers such that sensor 123 also hits consumer number 5.

aricart: @ripienaar the example is talking about summations, etc. For the "partitioning" problem, the bind to the subject must be an explicit thing - ie the list of subjects the consumer is interested in, must be augmented. - If the client is no longer available, the server cannot directly forward to any random consumer. IMHO this kind of partitioning will result in a different stream being created.

ripienaar: It's a class of problem, dont think how you could solve the summarization problem diferently.  Think how the summarization problem and the orders cancel before orders new problem while horizontally scaled can be solved by the same underlying assist from the server.  That's what this ADR is about.

Someone/something might decide a scale down event from 10 groups to 5 should happen and then yes, absolutely, we can decide another consumer can process the messages.  The critical promise is that all messages - for sensor 123, order 123, job logs related to job 123 - are processed in order.  That dimension is what's important.

ripienaar:
Closing in favor of https://github.com/nats-io/nats-architecture-and-design/pull/279 which is a early step in this direction

PR #41, opened 2021-08-07
kv-subject-name wildcard

scottf:
Use the cli command `nats kv add foo`, then do `nats s info KV_foo`,  the subject is `.*` not `.>` so assuming the docs should match the cli.

```
>nats s info KV_foo
Information for Stream KV_foo created 2021-08-07T16:49:16-04:00

Configuration:

             Subjects: $KV.foo.*
```

### DIFF
--- a/adr/ADR-8.md
+++ b/adr/ADR-8.md
@@ -193,7 +193,7 @@ Here is a full example of the `CONFIGURATION` bucket:
 {
   "name": "KV_CONFIGURATION",
   "subjects": [
-    "$KV.CONFIGURATION.>"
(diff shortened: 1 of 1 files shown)


ripienaar:
This being * breaks the entire ADR. It’s not a typo, much shown here simply won’t work with this change. 

I suspect your cli is a bit out of date

ripienaar:
It used to be a * - and old cli did that - this ADR and feature is in flux as we build and seek feedback. 

Hence no issue yet to build it into any clients.

scottf:
ok yeah the newest cli makes it `.>`

PR #42, opened 2021-08-11
JetStream subscribe workflow

kozlovic:
Signed-off-by: Ivan Kozlovic <ivan@synadia.com>


wallyqs (approved):
LGTM

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+### Creation
+
+The library should have API(s) that allow(s) creation of a JetStream subscription. Depending on the language, there may be various type of subscriptions (like Channel based in Go), or with various options such as Queue subscriptions, etc... There can be different APIs or one with a set of options.
+
+#### Configuration
+
+When creating a subscription, the user can provide several things:
+
+- A subject (see meaning of this below)
```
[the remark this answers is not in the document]
kozlovic: Will do..

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+The request is then sent to `<JS prefix>.CONSUMER.DURABLE.CREATE.<stream name>.<durable name>` for a durable subscription or `<JS prefix>.CONSUMER.CREATE.<stream name>` for an ephemeral.
+
+If the result of this request is "OK", it means that the server has created the JS consumer and uses the deliver subject the NATS subscription is created on. For ephemeral, the consumer name will be saved from the creation request's response, so that the ephemeral JS consumer can be deleted on Drain and Unsubscribe.
+
+If the result indicates that the consumer already exists, then it means that there is an active durable consumer, with some NATS subscription attached and so we should fail this, *unless* a Queue name was provided. The client/server currently does not use the concept of a queue group in the JS consumer, but if the server is enhanced, as part of getting the consumer information, the library will be able to determine if the consumer that already exists is part of the same group, in which case it is fine, or not, in which case an error should be returned.
```
kozlovic, line 95: @derekcollison I wonder if AddConsumer returns "already exists", is there not the case where we do PullSubscribe() and in that case, it is ok (but we should again verify that the returned consumer has no deliver subject if user wants PullSubscribe), and so it is an error only if it is push (deliver subject is specified) and there is no queue name?

Also, if the client were to send queue name when creating consumer, then server could send different error than "already exists" to make sure that the library fails if no queue name or queue name does not match what exists. This way, without having to update the client, the code will properly return an error.

scottf:
On one hand it's good to have a "managed" functionality; heartbeats / flow control / delete all handled for the user. Maybe this is how it will or should most often be used. But there will certain be users who want their own granular control of things, sbux comes to mind.
I'm not saying don't make the client smart so the user does not have to be, I'm just saying make sure the user has the option to use basic functionality on their own and make sure they can opt out of things like deleting a consumer.
Ordered Consumer is a good example. Pull mode Fetch. They are macro or managed functionality that package basic functionality in a useful way. But for instance always handling flow control and heartbeats for the user may not be appropriate for all clients or use cases.
Again, it's a matter of providing defaults and options.

scottf:
Regarding SemVer won't have to change, I would asses this on a client by client basis. API does not strictly mean only method signatures and interface changes, it also encompasses changes that would alter behavior that a user relies on. I'm not saying it will absolutely change for a given client, I'm suggesting that this is yet to be determined.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+Once the normal unsubscribe processing is done, and for a JS subscription that has been marked as needing to delete the JS consumer, the library should send a "DeleteConsumer" request and return the error if any.
+
+#### Drain
+
+The deletion of the JS consumer need to be delayed passed the point when the subscription is fully drained and removed from the connection. Since this is an asynchronous process, if the deletion of the JS consumer fail, an error will be pushed to the asynchronous error callback.
+
+## Consequences
+
+Some libraries have already been released with a workflow that is different and will need to align to the above design, however no public API signature is believed to need change, so it should not cause any SimVer issue.
```
scottf, line 137: I would asses this on a client by client basis. API does not strictly mean only method signatures and interface changes, it also encompasses changes that would alter behavior that a user relies on. I'm not saying it will absolutely change for a given client, I'm suggesting that this is yet to be determined.

scottf (changes requested):

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+}
+```
+
+#### Stream name
+
+When no stream name is specified, the library will use the subject provided as a way to find out which stream this subscription is for. A request is sent to the server on the `<JS prefix>.STREAM.NAMES` subject with a JSON content `{"subject":"<subject here>"}`. If the response (a list of stream names) is positive and contains a single entry, then the library will use this stream name, otherwise an error indicating that there is no matching stream name should be returned.
+
+#### Consumer name
+
+When no consumer name is specified, but a configuration is provided, the consumer will be the `Durable` field of the `ConsumerConfig` object.
```
scottf, line 60: Change requested:
A consumer name can be specified directly or as the `Durable` field of the `ConsumerConfig`. A value supplied directly takes precedence over the value supplied in the configuration if both are supplied.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+Note if the user provides a `Durable` in the configuration, this is treated differently that if the user provides a consumer to the subscribe API. Providing the stream and consumer name through the API makes the subscription being "bound", that is, no attempt to create a JS consumer will be done because the user expects the stream and consumer to already exists.
```
scottf, line 62: Change requested:
There is a separate API call `Bind` which requires a stream name and a durable consumer name. When binding, the subscription will be created with the assumption that both the stream and the consumer already exist.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+So if the consumer is not provided but set to the `ConsumerConfig.Durable` field, this means that if the durable exists, it will be attached to it, but if it doesn't, then the library will try to create this durable.
```
scottf, line 64: Requested to remove this.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+After this step, the library has a stream name and a consumer name, the latter possibly empty. If a consumer name is provided, the library will lookup the consumer info based on the pair (stream, consumer). If a consumer info is returned, some checks should be made so that the user provided `ConsumerConfig` matches what is returned by the server. Namely:
+
+- The `FilterSubject` matches the subject passed to the "Subscribe" API, if not return an error indicating the subject mismatch
+- If the user was trying to create a Pull subscription, but `DeliverSubject` is not empty, then return an error indicating that user can't create a pull subscription for a push based JS consumer
```
scottf, line 69: Existing behavior is to ignore (do not send in creation) deliver subject when creating a Pull consumer.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+- The opposite: if the user was not creating a Pull subscription but the `DeliverSubject` is empty, then return an error indicating that a pull susbcription is required
```
scottf, line 70: Existing behavior is to provide an inbox when deliver subject is not provided.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+
+If no error is found, then the internal NATS subscription should be created on the returned `DeliverSubject`.
+
+If there was a consumer lookup but an error occurs, the behavior depends on few things:
+
+- If the error is a "not found" (that is, the library got a response indicating that the said consumer does not exists):
+    - If the subscription is "bound" (stream and consumer names were provided), then unless this is a "pull" subscription, the error is returned.
```
scottf, line 77: bind is valid for both pull and push. If bind is requested and the consumer for the stream/durable is not found, it is an error.

adr/ADR-15.md
```diff
@@ -0,0 +1,137 @@
...
+- The consumer name, really a durable name for when the JS consumer already exists
+- A Queue name if this subscription is meant to be a queue subscription
+- A JS consumer configuration (see below), if the library is expected to create the JS consumer
+- Indication if this is a "Pull" subscription or not
+
+At the time of this writing, the consumer configuration object looks like this:
+
+```go
+type ConsumerConfig struct {
+	Durable         string        `json:"durable_name,omitempty"`
```
scottf, line 35: New field after Durable
```
	Description     string        `json:"description,omitempty"`
```

scottf (changes requested):

adr/ADR-15.md
kozlovic: I could rephrase, but I think it is important to distinguish between a `Bind` that will never attempt to create the JS consumer and a `Durable` that if not found will be created. So how would you state that?

adr/ADR-15.md
kozlovic: These checks are for when the consumer info is received (lookup returned an object) and we just ensure that the existing config matches what the user has provided. So yes, a pull subscribe will not set a deliver subject, but that is at the step when we create the JS consumer, here we are at the step where we have a consumer name and look it up.

adr/ADR-15.md
kozlovic: Same comment than above. This section describes checks that are done when we get a consumer, not when we create it.

adr/ADR-15.md
```diff
@@ -0,0 +1,138 @@
...
+
+If the result of this request is "OK", it means that the server has created the JS consumer and uses the deliver subject the NATS subscription is created on. For ephemeral, the consumer name will be saved from the creation request's response, so that the ephemeral JS consumer can be deleted on Drain and Unsubscribe.
+
+If the result indicates that the consumer already exists, then it means that there is an active durable consumer, with some NATS subscription attached and so we should fail this, *unless* a Queue name was provided. The client/server currently does not use the concept of a queue group in the JS consumer, but if the server is enhanced, as part of getting the consumer information, the library will be able to determine if the consumer that already exists is part of the same group, in which case it is fine, or not, in which case an error should be returned.
+
+When the consumer already exists, the NATS queue subscription that was created prior to the "AddConsumer" call needs to be destroyed and replaced with the new NATS queue subscription on the consumer info's `DeliverSubject`.
+
+### Runtime
+
+#### Heartbeats
```
ripienaar, line 102: we need to update for recent discussions that we should always handle them even if we think they shouldnt be coming or we jsut dont know.

Maybe worth mentioning about FC with details in the next section

adr/ADR-15.md
```diff
@@ -0,0 +1,138 @@
...
+
+The server includes in the heartbeats a header to indicate if the consumer is stalled and a subject to send an empty message to. When that happens, the library need to respond to this message, regardless of its internal state of flow control.
+
+Without this, it can be that a consumer remains stalled, so having both heartbeats and flow control configured seem like necessary, so we may make that a misconfiguration if flow control is configured without heartbeat.
+
+Again, Ordered consumers should be favored.
+
+### Unsubscribe and Drain
+
+When a subscription is unsubscribed or drained, if the library created the JS consumer, (or it already existed and this is a queue member added to the group), at the end of those calls the JS consumer should be deleted.
```
ripienaar, line 126: If we joined an already-existing consumer as a queue member, shouldnt we leave it alone and not delete?

ripienaar (changes requested):
Might be worth listing seperately all the APIs this interact with - I know you call them out but a nice reference would be nice

scottf:
I proposed some significant changes to this which cannot be captured in review comments. Please see https://github.com/nats-io/nats-architecture-and-design/pull/46

adr/ADR-15.md
```diff
@@ -0,0 +1,231 @@
...
+
+The library should keep track of all messages which reply subject start with `$JS.ACK` since they contain information about the stream and consumer sequence, etc..
+
+When the library receives a message that is found to be a control message (no payload and containst the "Status" header), if the header is "100 Idle Heartbeat", then the library should check if there is any gap detected between what was the last tracked message meta information and the content of the heartbeat. If a gap is detected, an asynchronous error should be reported indicating what the gap is.
+
+Ordered consumers (should have its own ADR) will handle all that for the application, recreating the subscription as needed at the right stream sequence.
+
+**Note**: In most libraries the hearbeats messages are never presented to the user. Those are handled internally.
+
+#### JS Ack
```
aricart, line 150: It would be more clear if we call this the `DeliveryInfo` - this is the data encoded into the reply subject of a JetStream message sent to a subscriber.

kozlovic: Added a bit of verbiage to clarifies that we are talking about the ACK/delivery information for a message sent by the server to a subscriber.

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+When the server sends a message to a subscriber, and if there is an AckPolicy specified other than "AckNone", then the message will have a reply subject that the library is supposed to use in order to acknowledge the message. This subject also encodes some delivery information, such as stream and consumer name, stream sequence, etc..
+
+Until now, ACK reply subject contained 9 tokens, with this layout:
+```
+$JS.ACK.<stream name>.<consumer name>.<num delivered>.<stream sequence>.<consumer sequence>.<timestamp>.<num pending>
+```
+
+This is going to change with newer servers where the number of tokens will be 10 or 11, with the total tokens being:
+```
+$JS.ACK.[<domain>.]<account hash>.<stream name>.<consumer name>.<num delivered>.<stream sequence>.<consumer sequence>.<timestamp>.<num pending>
```
ripienaar, line 161: I asked this on slack but worth capturing here, when we modified these why did we chose to prepend the domain and account instead of appending?

If we appended the handling of this in clients would be significantly simpler and if we make a change now so we are not strict in the token count of the subject, any future work added to the end of these subjects would essentially remain gracefully backward compat.  Old clients will work they just won't know about the new data.

I imagine it might be because the server might want to do a single subscribe against `$JS.ACK.domain.>` or something, so wanted to find out.

Additionally how long is the account hash? Hopefully its not a 32byte hash like md5 or something, thats a LOT of bytes over time.

kozlovic: Nothing against that, and Matthias is having a hard look at the moment to make sure that the decision is not undone few days later.
Still, for some libraries, adding at the end would still break anyway. For instance, the released Go client v1.11.0 is checking for the number of tokens: https://github.com/nats-io/nats.go/blob/3b1f6fcc1e1014c838036367494c3012523166b0/js.go#L1976

ripienaar: Indeed, we have libraries that check the length, so some needs to change, I just want to be sure that if we change we start on an approach that will inform future changes as well

kozlovic: Completely agree with that.

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+
+Again, Ordered consumers should be favored.
+
+**Note**: In most libraries the flow control messages are never presented to the user. Those are handled internally.
+
+### Unsubscribe and Drain
+
+When a subscription is unsubscribed or drained, if the library created the JS consumer (when it called "AddConsumer" and got an OK), the JetStream consumer is deleted at the end of those calls.
+
+**Note** I think it is problematic for queue subscriptions since the first member to start will create the JS consumer (in cases where the JS consumer does not already exists), but will be marked as needing to delete the JS consumer on Unsubscribe()/Drain(), while other members may have attached to the same JS consumer.
```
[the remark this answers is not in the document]
derekcollison: We IIRC are aware of this and wanted to introduce the notion of a force option when updating or deleting and switch normal behavior to blocking some of these when we know we are bound, like queue subscriber situation above.

We could move forward now and say if you try to delete a bound push consumer that will fail (bound == active). And that to force it we need to add in the force option.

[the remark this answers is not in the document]
kozlovic: > We could move forward now and say if you try to delete a bound push consumer that will fail (bound == active).

That means in the server, if I understand correctly. So this means that we keep the client as it is now (calling js.DeleteConsumer if the library called AddConsumer(), regardless of the type of JS consumer: ephemeral/durable/queue durable)

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+
+The library should have API(s) that allow(s) creation of a JetStream subscription. Depending on the language, there may be various type of subscriptions (like Channel based in Go), or with various options such as Queue subscriptions, etc... There can be different APIs or one with a set of options.
+
+#### Configuration
+
+When creating a subscription, the user can provide several things:
+
+- A subject that is in some cases required (see "Stream name" section below)
+- An optional stream name
+- An optional consumer name
```
scottf, line 26: can this be `An optional durable consumer name`

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+- An optional JS consumer configuration (see below), if the library is expected to create the JS consumer
```
scottf, line 27: A JS consumer could be created even though consumer config is not supplied. There are defaults for the consumer config that would be assumed. This would suggest removing `, if the library is expected to create the JS consumer`

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+- A Queue name if this subscription is meant to be a queue subscription
+- Indication if this is a "Pull" subscription or not
```
scottf, line 29: can this be `Indication if this is a "Push" or "Pull" subscription` ? I understand that it's binary, but I think more description is valuable.

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+- An optional way of indicating the the given subject will be used to create the internal subscription, by-passing any lookup/consumer creation. For the purpose of this document, let's call this option `SubjectIsDelivery`.
```
scottf, line 30: This is confusing. By the last sentence is this something other than 'bind'?
Also where is the `Indication if this is a "Bind"` part. Bind was a previously agreed upon term and has already been coded (although not released)

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+
+Some misconfiguration should be checked by the subscribe API and return an error outright.
+
```
scottf, line 33: what about `Handlers for receiving messages asynchronously, whatever that means in the library language. `

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+	MaxAckPending   int           `json:"max_ack_pending,omitempty"`
+	MaxDeliver      int           `json:"max_deliver,omitempty"`
+	MaxWaiting      int           `json:"max_waiting,omitempty"`
+	OptStartSeq     uint64        `json:"opt_start_seq,omitempty"`
+	OptStartTime    *time.Time    `json:"opt_start_time,omitempty"`
+	RateLimit       uint64        `json:"rate_limit_bps,omitempty"` // Bits per sec
+	ReplayPolicy    ReplayPolicy  `json:"replay_policy"`
+	SampleFrequency string        `json:"sample_freq,omitempty"`
+}
+```
```
scottf, line 64: This seemed like a good place to note errors. Maybe they are noted further below, or are they not going to be mentioned?
```
OptStartSeq     uint64        `json:"opt_start_seq,omitempty"`    // server returns error if provided when Deliver Policy is not by_start_sequence
OptStartTime    *time.Time    `json:"opt_start_time,omitempty"`   // server returns error if provided when Deliver Policy is not by_start_time 
MaxWaiting      int           `json:"max_waiting,omitempty"`      // server returns error if provided if a DeliverSubject is provided
```

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+
+#### Stream name
```
scottf, line 66: Can you start with a section describing the `SubjectIsDelivery` option? I have no context at this point to what it means.

scottf (changes requested):

adr/ADR-15.md
kozlovic: No, it is a consumer name because it does not have to be a durable. For instance, in Go, if you use Bind(stream, consumer) you specify a consumer name, which can be in theory an ephemeral (non durable).

adr/ADR-15.md
kozlovic: Will remove.

adr/ADR-15.md
kozlovic: will change.

adr/ADR-15.md
kozlovic: Bind is when you specify a stream and consumer name. It has been done in Go already in the already released v1.11.0.

kozlovic: Yes, this option is to say that the library will go directly create the NATS subscription on the subject passed to the "subscribe" API call, and won't do any creation of consumer/lookup/or any check.

adr/ADR-15.md
kozlovic: Ok, can add that, but you have sync subscriptions too, and what if there is a library that does not have the concept of async, I don't know. But sure, I can put that on.

adr/ADR-15.md
kozlovic: Will add

adr/ADR-15.md
kozlovic: I have created a section before "Stream name" that describes this option and gives some details. Hope this clarifies it.

adr/ADR-15.md
```diff
@@ -0,0 +1,242 @@
...
+
+For push consumers, if the `DeliverSubject` is not specified, the library will pick an inbox name. For pull consumers, `DeliverSubject` needs to be left blank.
+
+The library will set the `FilterSubject` to the user provided subject, ensure that `AckPolicy` is set and possibly some other fields such as `MaxAckPending` and set the `DeliverGroup` to the queue name if the subscribe call is for a queue subscription.
+
+The request is then sent to `<JS prefix>.CONSUMER.DURABLE.CREATE.<stream name>.<durable name>` for a durable subscription or `<JS prefix>.CONSUMER.CREATE.<stream name>` for an ephemeral.
+
+If the result of this request is "OK", it means that the server has created the JS consumer and uses the deliver subject the NATS subscription is created on. At this point, the jetstream subscription will keep track of the fact that it did create the JetStream consumer and will delete it on Unsubscribe() or Drain(). For ephemerals, the consumer name will be saved from the creation request's response, since it was not available beforehand.
```
aricart, line 135: create consumer request returns the consumer config?

adr/ADR-15.md
aricart: @ripienaar @kozlovic the reason to push it forward is to simplify the export/import subject, as that information can make it to the front of the subscription. ie `$JS.ACK.<domain>.<account>.>`

aricart (approved):
LGTM

adr/ADR-15.md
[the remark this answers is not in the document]
derekcollison: We can add that behavior and the additional force option to those calls in the server for sure.

adr/ADR-15.md
kozlovic: Correct. A new token will be added at the end and recommendations maybe to have the library not check the length will be added (if we add tokens at the end in the future).

adr/ADR-15.md
kozlovic: This new option is removed.

wallyqs (approved):

adr/ADR-15.md
```diff
@@ -0,0 +1,233 @@
...
+
+The account hash is not used by the client at this time, only used for routing, same for the last token.
+
+It is recommended that library no longer report an error if there are more than 12 tokens, so that new tokens may be added, especially if their purpose is solely for routing and have no impact on the library itself.
+
+#### Flow control
+
+When a JetStream consumer has flow control enabled (not applicable to pull consumers), the library may receive a control message that indicates that this is a flow control and should respond to the provided subject when the current pending message has been processed.
+
+That is, suppose messages are received on a connection and pushed to a JetStream subscription, and there are currently 1000 pending messages. When the connection received a JetStream flow control status message for a given subscription, it should mark that the subscription should send an empty message to the given Flow Control subject (which is the reply subject of the incoming control message) when the library has dispatched that 1000's message in the subscription's pending list.
```
matthiashanel, line 180: Just to make sure this is not happening: The format or the number of tokens of the Fc message is not to be inspected.

kozlovic: Added a note.

adr/ADR-15.md
```diff
@@ -0,0 +1,235 @@
...
+	SampleFrequency string        `json:"sample_freq,omitempty"`
+}
+```
+
+#### Stream name
+
+When no stream is specified, the `subject` to the subscribe API calls is required. An error will be returned if it is not the case.
+The library will use the subject provided as a way to find out which stream this subscription is for. A request is sent to the server on the `<JS prefix>.STREAM.NAMES` subject with a JSON content `{"subject":"<subject here>"}`. If the response (a list of stream names) is positive and contains a single entry, then the library will use this stream name, otherwise an error indicating that there is no matching stream name should be returned.
+
+If the stream name is specified, then the `subject` is possibly not needed.
```
scottf, line 69: should clients assume a subject equal to the stream name?

adr/ADR-15.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Consumer name
+
+When a consumer name is specified, it indicates the the library that the intent is to bind to an existing JS Consumer. The library will lookup the consumer from the server and get a `ConsumerInfo`. If the lookup fails, it is considered an error, unless the subscription is for a "pull subscriber", in which case the library still proceeds with the JS subscription.
```
scottf, line 73: 1. can you change the word 'bind' to something like 'use' or 're-use'
2. Also can you use the word durable in this context with 'consumer name'

adr/ADR-15.md
```diff
@@ -0,0 +1,235 @@
...
+
+If the user attempts to create a queue subscription, a consumer name or durable needs to be specified. The common pattern error that we noticed user doing was this:
+
+```go
+member1, _ := js.QueueSubscribe("foo", "bar")
+member2, _ := js.QueueSubscribe("foo", "bar")
+```
+And then report that both members were receiving the same message.
+
+This is because each subscription call would create their own ephemeral JetStream consumer. We could break the API and force the user to specify a consumer name. Instead, for the Go client we have taken the approach to describe that if no consumer name is provided (in Go with `nats.Bind(stream, consumer)`) nor a durable (in Go with `nats.Durable(durableName)`) then the library would use the queue name as the durable name.
```
scottf, line 85: With the new `deliver_group` field in the consumer config, the queue name should be used as the deliver group. is a durable name even required when a deliver group is specified, or can ephemerals participate in a deliver group? 
If ephemerals do not participate in a user group, I suppose there is no reason why the durable name cannot be the same as the deliver group name, but I would check with the server to see if this is acceptable.
Also, should we retire the queue concept for deliver group concept? I think queue group is actually a confusing name to begin with because queue really means something else. I know, naming is hard.

adr/ADR-15.md
kozlovic: Now that we have removed the SubjectIsDelivery, things change a bit, so I may update this. If it is a pull subscription, if the stream is specified, this subject will really not be needed if the consumer already exists. But for non pull, or when creating the JS consumer, it is needed because we set FilterSubject with the given subject, or, for the lookup we verify that if a filter subject is set, it matches the user given subject.
So I will see how I update that.

adr/ADR-15.md
kozlovic: 1. ok
2. Sorry but no. You can reuse a consumer that is non-durable as long as you know its name. Sure it won't likely be the default behavior, but assuming that a consumer name means durable is misleading.

adr/ADR-15.md
kozlovic: The confusion you have is that an ephemeral consumer is not a durable, but it is still a consumer and it has a name. (but happen - as of now - to be selected by the user when it is created). This is why I describe above that I can't change "consumer name" to "durable" everywhere because it is not correct.
(for information: a durable is replicated across server nodes, while an ephemeral is always "R1", which makes it faster).
As explained earlier, one *could* create a queue group (call many instances of js.QueueSubscribe() with the same queue name) for an ephemeral JS consumer, using `nats.Bind(stream, consumer)` option, in that case without `nats.Durable()` option. That is, an admin could create an ephemeral (with a given deliver group) and give that name to application owners. Again, unlikely, but possible.

ripienaar (approved):
LGTM

aricart (commented):
LGTM - I have implemented the queue and server pub ack metadata parsing already on the client. the other changes will require a more careful look to insure that it matches (most looks right to me), but requires a detailed code review.

PR #45, opened 2021-08-12
[adding] ADR-17 Ordered Consumer

scottf:
See the rendered version here: 
https://github.com/nats-io/nats-architecture-and-design/blob/adr-17-ordered-consumer/adr/ADR-17.md


adr/ADR-17.md
```diff
@@ -0,0 +1,48 @@
...
+
+The subscription is not allowed with queues/deliver groups.
+
+### Consumer Configuration Checks
+
+The user can provide a consumer configuration but it must be validated. Error when creating if validation fails.
+
+Checks:
+
+- durable_name: can not be provided
```
kozlovic, line 40: You removed the deliver subject check, but it was correct to say that it is not allowed. In go, we have this check:
```
// No deliver subject, we pick our own.
if o.cfg.DeliverSubject != _EMPTY_ {
    return nil, fmt.Errorf("nats: deliver subject can not be set for an ordered consumer")
}
```
The reason is that we want the ephemeral created to have a unique (inbox) deliver subject, which helps when doing a "reset" where the old one is stopped and new one created. Having the same subject may be an issue, I think.

kozlovic (changes requested):

adr/ADR-17.md
kozlovic: PS: I missed that one in the current C PR, so will add to it.

scottf: ok I can add that instruction back in.

PR #48, opened 2021-08-17
[update] ADR-6 add bucket and key naming conventions

scottf:


aricart:
I was just looking at this, as I was updating the JavaScript implementation. The realization that I have on this is that we can validate the key names (currently in go/javascript it allows `-`, `/`, `=`, alpha, digits, and `.` so long as they are not a start or end character). Buckets are the same but `.` is not allowed anywhere.

```javascript
const validKeyRe = /^[-/=.\w]+$/;
const validBucketRe = /^[-\w]+$/;
```

However quickly this becomes an issue because the number of characters that probably should be allowed is much larger (as Scott is pointing out). - So the quick realization is that the key encoding function (and there must be one), is really the correct strategy or unexpected errors will creep up, as key values from unknown sources are now subject to a runtime error.

This means that all keys should, split on "." in case they gave some compound key, extract the wildcards as they are, and encode all else, rejoining with a ".":

```javascript
  function encodeKey(key: string): string {
    const chunks: string[] = [];
    for (const t of key.split(".")) {
      switch (t) {
        case ">":
        case "*":
          chunks.push(t);
          break;
        default:
          chunks.push(this.codec.key.encode(t));
          break;
      }
    }
    return chunks.join(".");
  }
```

ripienaar:
Your `.` in the key reflex needs to be escaped. 

I specifically restricted these to maximise the possibility of being able to integrate with other tools.

ripienaar:
I have this for encoding keys 

```
func (j *jetStreamStorage) encodeKey(key string) (string, error) {
	res := []string{}
	for _, t := range strings.Split(key, ".") {
		if t == ">" || t == "*" {
			res = append(res, t)
			continue
		}

		et, err := j.encode([]byte(t))
		if err != nil {
			return "", err
		}

		res = append(res, string(et))
	}

	return strings.Join(res, "."), nil
}
```

adr/ADR-6.md
```diff
@@ -53,4 +56,6 @@ durable-name       = term
 js-internal-prefix = dollar (prefix dot)+
 js-user-prefix     = (prefix dot)+
 account-name       = (filename-safe)+ maximum 255 characters
+kv-bucket-name     = restricted-term
+kv-key-name        = term
```
ripienaar, line 60: key name is not `term` at all.

`.` is allowed but not at start, all the various punctuations in 33-126 ascii is not allowed etc. It's `[-/_=\.a-zA-Z0-9]` and may not start with `.`

scottf: okay was not sure, figure you would catch that in review. I will update.

adr/ADR-6.md
```diff
@@ -56,6 +58,6 @@ durable-name       = term
 js-internal-prefix = dollar (prefix dot)+
 js-user-prefix     = (prefix dot)+
 account-name       = (filename-safe)+ maximum 255 characters
+kv-key-name        = limited-term (dot limited-term)*
```
ripienaar, line 61: ```-/_=.a-zA-Z0-9```

scottf: right backslash was escape, will fix

ripienaar: they also may not start or end with `.`

```
Valid keys are \A[-/_=\.a-zA-Z0-9]+\z, additionally they may not start or end in ., after encoding
```

Please review ADR-8

ripienaar (approved):
LGTM

PR #54, opened 2021-09-20
[change] ADR-15 Automatic Status Management

scottf:


ripienaar:
Shouldn’t this be an addition to the existing document detailing how subscribe work? Seems like a perfect fit in that existing ADR

adr/ADR-19.md
```diff
@@ -0,0 +1,35 @@
...
+
+ASM should be the default client behavior. Optionally, clients can allow users to choose to handle the status messages themself.
+
+## General
+
+Intercept all status messages, never pass to user. If there is some exceptional reason to notify or escalate some state to the user, user a language specific paradigm such as an exception.
+
+## Flow Control and Heartbeat
+
+1. Automatically respond to flow control messages as soon as it is observed in the normal processing of next message. Escalate to user when unable to respond to the flow control.
```
kozlovic, line 22: Maybe I misread what you wrote here, but when receiving a Flow Control, the library needs to respond *only* after all pending messages have been processed. Otherwise, this defeats the point of FC.

adr/ADR-19.md
```diff
@@ -0,0 +1,35 @@
...
+1. Have an alarm timer that escalates an issue to the user when heartbeats have been missed. Optionally provide a way to specify the period and threshold for an alarm. Default period should equal the heartbeat time and default threshold should be 3 (???)
+
+## Pull
+
+1. Automatically capture 404 and 408 statuses. Those messages are not considered in fetch or iterate counts
+
+## Consequences
+
+Existing client behavior may be released that the default or only behavior is to pass status messages to the user. 
+Changning to ASM as the default, while not technically breaking, should be documented in release notes and readmes.
```
kozlovic, line 34: typo: Changing instead of Changning

kozlovic (changes requested):

adr/ADR-19.md
scottf: "as soon as it is observed"

scottf: I am suggesting that it's processed as the user would normally get the message. I would argue against processing it when is received over the network. If the user isn't processing messages, it makes sense that they are not ready to resume getting messages from the server.

adr/ADR-19.md
scottf: fixed

adr/ADR-19.md
scottf: Back to my example. I get 350 messages, then I get the FC, but the server has already sent me another 300 or so messages and they are buffered. I would not hold the fc until the buffer is empty.

kozlovic: Well no, the server will not send you another 300 messages if it has sent an FC to which the library did not respond to. But as long as we are clear that the library should not simply reply to an FC message when it gets it, but instead when all previous messages have been delivered, then it is ok.

scottf: I think we are saying the same thing. There is a difference between the client buffering the messages the server has sent and the user asking for it. It doesn't matter if the FC is in the buffer, it is only responded to when the user asks for the next message and it turns out to be the fc.

kozlovic: Right

adr/ADR-15.md
```diff
@@ -127,71 +127,114 @@ If the result indicates that the consumer already exists, then it means that the
...
-Without this, it can be that a consumer remains stalled, so having both heartbeats and flow control configured seem like necessary, so we may make that a misconfiguration if flow control is configured without heartbeat.
+It's possible that because of the idle heartbeat duration, that both flow control and multiple heartbeat messages contain the
+flow control subject. It is only required to respond to the specific subject once, so it is suggested that clients track 
+the last flow control subject responded to avoid replying multiple times for the same flow control. The server will ignore duplicates, it is just unnecessary traffic.  
 
-Again, Ordered consumers should be favored.
+#### Pull Mode Statuses
 
-**Note**: In most libraries the flow control messages are never presented to the user. Those are handled internally.
+In pull mode, 404 and 408 statuses should not be surfaced to the user or considered in fetch or iterate counts.
```
wallyqs, line 237: Any status in general, there is also 409... then also 503 becomes an error etc...

scottf: @wallyqs What are 409 and 503 for?

ripienaar: 409 - doing conflicting things like trying to pull from a push consumer, exceeding max ack pending etc
503 - no responder, like if you try to reply to fc and the subscribe went away

scottf: So it seems like 409 and 503 should not be passed, but an error should be raised (not passed) to the user, unlike 404 and 408 which can be discarded. Updating the doc. Should we assume that unknown statuses should be treated as errors?

kozlovic (approved):
LGTM

scottf:
Added notes regarding queues and flow control/heartbeat and also regarding queues and gap detections. TLDR; the don't work together, handle these cases

kozlovic (approved):
LGTM

PR #55, opened 2021-09-27
[change] ADR-8 KV differentiate get and get entry

scottf:


adr/ADR-8.md
```diff
@@ -116,8 +116,11 @@ The interface here is a guide of what should function in read-only mode.
 ```go
 // RoKV is a read-only interface to a single key-value store bucket
 type RoKV interface {
-	// Get gets a key from the store
-	Get(key string) (Entry, error)
+	// Get gets a key value from the store
+	Get(key string) ([]byte, error)
+
+        // Get gets a key entry from the store
+	GetEntry(key string) (Entry, error)
```
ripienaar, line 123: Would rather keep `Get` as is and document that languages are free to add additional get methods like `GetBytes()` and `GetInt()` etc

scottf: So the getInt is a variety of get value, not get entry. I felt like they were two distinct behaviors, the get value the most common use, but getEntry certainly has value. It just seems right to specify both here.

ripienaar: I'd prefer we keep Get() as the one we can all uniformly do and then perhaps show an example that languages can add GetInt() or GetValue() or GetUint64() or whatever.

scottf: It's clear that the majority of use cases for `get` is to get the value. getString/getLong are overloads of getting the value, not overloads of getting the entry. This is the same idea as with headers, that although we support a list of values, getting the key by default returns the first one value since there is almost always only one value. Also, again, yes there are uses for get entry, but the majority of applications the majority of the time want the value.

kozlovic: If we agree that we should have all implementations have at the minimum a Get that returns a full entry and optionally have APIs that return specific values/types, the I would agree with RI that we should have Get() return the entry. Again, if it is the minimal API set, I would rather have Get() than GetEntry().

aricart: For uniformity, I think that Get under the current KV should return a KvEntry, Your entry implementation can get the value in the format you want or folks can wrap the entry to decode into a specific type. This means that the public API is as compact as possible (byte-oriented as NATS is).

scottf: I've changed my opinion and am proposing a change.

kozlovic: Well, I have state my preference, and I align with RI: I would have Get() return the entry. You should not have add GetValue() here, what you should do is state outside of this interface definition that libraries are welcome to add specific getters if desired.

kozlovic: I see that in Java you already have getEntry(), but you have marked the feature/APIs as Beta, so you could still change I suppose.

scottf: This is a Key Value store, not a Key Entry store

scottf: @aricart Uniform to what? This is not uniform with headers where the underlying value is a list, but get returns the first value in the list _because_ 99% of the time there is only one value

ripienaar: There is ample prior art to support this design. 

https://pkg.go.dev/github.com/hashicorp/consul/api#KV.Get

scottf: @kozlovic Why are defining a "minimum" when we need client parity. 99% of the use cases, the user wants a value, not the entry. To get a value from an entry, I have to first check if I got an entry back and then if I did, that I can get it's value.

ripienaar: And here is the etcd interface

https://pkg.go.dev/github.com/hashicorp/consul/api#KV.Get

which is even fatter as they support ranged gets in Get()

so pretty much this is pretty normal design and these most widely used systems influenced the design of our interface.

scottf: The java already has `getValue` and `getEntry` defined (it's beta). I had forgotten this when I made the proposal which I have now changed to match that. For me to change `getEntry` to `get` is trivial. But I think it's confusing while having getValue and getEntry are unambiguous. Having to specify getValue as a "minimum" seems reasonable.

ripienaar: And here is kine - the one that Derek is keen to see mimicked. Also a fat value response. 

https://github.com/k3s-io/kine/blob/master/pkg/client/client.go#L23

It’s a pity you didn’t follow the ADR to begin with. My suggestion is that, given the consensus and ample prior art in the industry, you update the Java client to match the ADR.

scottf: I did follow what I consider the spirit of the ADR. I made the interfaces available to everyone for review quite a while ago. In addition, I split management functions off into a separate interface. I provide all the functionality in the ADR, it's simple and clear, it doesn't match exactly the division of functionality aka RoKV / KV as the ADR states. I considered the normal Java application user when developing.
In either case, I'm closing this PR without merging it. I'm leaving the Java interface as is, it makes sense and is unambiguous as to what the functionality is.

jnmoyne: FWIW in ActiveSpaces we went through the same discussion, had started with the Get returning the value but pretty quickly changed it to return an entry (if there was no value associated with the key then the returned value would be Null. Then from the Entry you could get the value (Tuple) and some of the metadata. 

Also Entry was used as input for the compare and set (aka compare and swap) (and I think a few others if memory serves me right despite my old age).

PR #56, opened 2021-09-27
document expect last sequence behavior

ripienaar:


adr/ADR-8.md
```diff
@@ -216,7 +216,11 @@ Writing a key to the bucket is a basic JetStream request.
 The KV key `auth.username` in the `CONFIGURATION` bucket is written sent, using a request, to `$KV.CONFIGURATION.auth.username`.
 
 To implement the feature that would accept a write only if the sequence of the current value of a key has a specific sequence
-we use the new `Nats-Expected-Last-Subject-Sequence` header.
+we use the new `Nats-Expected-Last-Subject-Sequence` header. Since a recent server version the special value `0` for this header
+would indicate that the message should only be accepted if it's the first message on a subject. This is purge aware, ie. if 
+a value is in and the subject is purged again a `0` value will be accepted.
+
+This can be implemented as a `PutOption` ie. `Put("x.y", []byte("hello world"), OnlyIfLastValueSequence(10))`.
```
ripienaar, line 223: Please suggest better name than `OnlyIfLastValueSequence`

scottf: I suppose this is intended to solve the create vs put problem but does not because what if a key's last operation is delete, a create should work. Maybe a purge should happen first. Besides the create scenario, I can't see a user using this at all. Maybe we just have to add create to the api

ripienaar: There are many things people do with KV around service discovery where either a specific thing should be standby/fiirst, where without updates expires should happen, where you should be able to iterate members in a set etc. 

All weird and wonderful things people do with KV (see Consul service registry), so you might not have used it in that case and indeed for the typical put some config stuff in a bucket its entirely useless, but being present in the API mean we can fix other problems too

There's a whole other world of things people build with KV that I am working on - service registries is one of the major reasons people run consul.

scottf: You could just name it to match the header name, i.e. `ExpectedLastSequence`

derekcollison, line 223: I kinda like Create() here tbh.

ripienaar:
Updated this to remove `Purge()` and add a `Purge(key)` with a note about it only being viable later once we have Roll-up feature in the server

ripienaar:
Updated this in light of discussions around API design.
Added RollUp based key purge design based on https://github.com/nats-io/nats-server/pull/2585

adr/ADR-8.md
```diff
@@ -254,9 +260,13 @@ place a new message in the subject for the key with a nil body and the header `K
 This preserves history and communicate to watchers, caches and gets that a delete operation should be handled - clear cache,
 return no key error etc.
 
-#### Purging a bucket
+#### Purging a key
+
+Purge is like delete but history is not preserved. This is achieved by publishing a message in the same manner as Delete using the
+`KV-Operation: DEL` header but adding the header `Nats-Rollup: sub` in addition.
```
scottf, line 266: Maybe make a note that this means do not simply do a straight subject purge, that was my original implementation which will be replaced

adr/ADR-8.md
```diff
@@ -318,9 +382,9 @@ While the cache is still warming up it's a pass through to the underlying backen
 Most operations are pass-through, except:
 
  * `Get()` will read local cache if ready (fully warmed), and a key is there
- * `Delete()` will evict the key from local cache and pass through
+ * `Delete()` and `Purge()` will evict the key from local cache and pass through
  * `Put()` will evict the key from local cache and pass through
- * `Destroy()` and `Purge()` will delete all messages from local cache and pass through
+ * `Destroy()` will delete all messages from local cache and pass through
```
scottf, line 387: I'm wondering if we should make local caches an option to the context. Local cache will require a watch running, which means we must have a background thread / async subscription. Maybe someone won't want or need this if they are know they are close to the server.

adr/ADR-8.md
scottf: I support the idea of having `Put` as the minimum but `Create` and `Update` can be coded easily enough with Get (Entry) first or put with sequence expectations. 
The problem I do see, is who decides (the user or us) whether Create is valid if the key was deleted as well as has has never existed. Similar with update. So that behavior would have to be defined.
Looking further down in the changes, this could be accomplished by the options versions of the put. The options need to be defined.

adr/ADR-8.md
```diff
@@ -293,6 +303,60 @@ all the known services we ask for all subjects below `$KV.SERVICES.nats.>`. A no
...
+
+So the overall consensus is that a `Entry` like entity should be returned, hard to say but indeed `Get(string) ([]byte,error)` would seem
+like a good default design, but in the face of massive evidence that this is just not the chosen design I picked returning a `Entry` as 
+the default behavior.  But leaving it open to languages to add additional helpers.
+
+Given that `get()` as a name isn't impossible in Java - see etcd for example - I think in the interest of harmonized client design we should
+use `get()` wherever possible augmented by other get helpers.
+
+Regarding `Put`, these other APIs do not tend to add other functions like `Create()` or `Update()`, they accept put options, like we do.
+Cases where they want to build for example a CAS wrapper around their KV they will write a wrapper with CAS specific function names and more,
```
scottf, line 353: I commented above on create/update. Redis and Memcache do support this. I think Mongo for instance does not.

adr/ADR-8.md
```diff
@@ -293,6 +303,60 @@ all the known services we ask for all subjects below `$KV.SERVICES.nats.>`. A no
...
 
+#### API Design notes
+
+The API here represents a minimum, languages can add local flavour to the API - for example one can add `PutUint64()` and `GetUint64()`
+in addition to the `Get()` and `Put()` defined here, but it's important that as far as possible all implementations strive to match
+the core API calls - `Get()`, `Put()` and to a lessor extend `Delete()` and `Purge()` and `Entry` - the rest, admin APIs, and supporting
+calls can be adjusted to be a natural fit in the language and design.  This is in line with existing efforts to harmonize 
+`Subscribe()`, `Publish()` and more.
+
+The design is based on extensive prior art in the industry. During development 20+ libraries in Go, Ruby, Python and Java were reviewed
```
scottf, line 314: What libraries? All etcd relative? Were cache libraries for Redis / Memchaced considered. NoSql databases?

adr/ADR-8.md
```diff
@@ -293,6 +303,60 @@ all the known services we ask for all subjects below `$KV.SERVICES.nats.>`. A no
...
+```go
+func (k *KV) Put(p *KVPair, q *WriteOptions) (*WriteMeta, error){}
+func (k *KV) Get(key string, q *QueryOptions) (*KVPair, *QueryMeta, error)
+```
+
+Here KVPair has properties like `Key`, various sequences, flags, timestamps etc and `Value []byte`.
+
+Dynanic language libraries like Python for example return maps, but there too the return value of a default `get()` operation is a hash
+with all metadata and value stored in `Value` or similar.
+
```
scottf, line 328: hmm. Maybe I'm looking at the wrong thing: https://docs.python.org/3/tutorial/datastructures.html#dictionaries

scottf (changes requested):

adr/ADR-8.md
```diff
@@ -293,6 +303,60 @@ all the known services we ask for all subjects below `$KV.SERVICES.nats.>`. A no
...
+The official `jetcd` library has a `get()` that returns a future and an equally fat ranged response.
+
+Various Ruby etcd clients were reviewed, same design around `get` with a object returned.
+
+So the overall consensus is that a `Entry` like entity should be returned, hard to say but indeed `Get(string) ([]byte,error)` would seem
+like a good default design, but in the face of massive evidence that this is just not the chosen design I picked returning a `Entry` as 
+the default behavior.  But leaving it open to languages to add additional helpers.
+
+Given that `get()` as a name isn't impossible in Java - see etcd for example - I think in the interest of harmonized client design we should
```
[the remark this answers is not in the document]
ripienaar: I simply missed those issues, they are issues though. Would be great if we all can implement the same design.

adr/ADR-8.md
ripienaar: it's opt in

adr/ADR-8.md
ripienaar: I looked at a bunch, for the KV stores we tend to hear from, redis a hot mess of millions of options so did not consider doing anything resembling it.

adr/ADR-8.md
```diff
@@ -293,6 +303,60 @@ all the known services we ask for all subjects below `$KV.SERVICES.nats.>`. A no
...
+```
+
+Here `GetResponse` is much fatter than we propose as it also supports ranged queries and so multiple values, each being like our Entry.
+
+The official `jetcd` library has a `get()` that returns a future and an equally fat ranged response.
+
+Various Ruby etcd clients were reviewed, same design around `get` with a object returned.
+
+So the overall consensus is that a `Entry` like entity should be returned, hard to say but indeed `Get(string) ([]byte,error)` would seem
+like a good default design, but in the face of massive evidence that this is just not the chosen design I picked returning a `Entry` as
```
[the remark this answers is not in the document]
ripienaar: Yet, one has to be chosen, we chose when the ADR was approved.  Just like you claim your interfaces has been out there for a month- the ADR has been there longer.  It's the exact same scenario.  Some choise has to be made.  Our process is the follow the ADR.

adr/ADR-8.md
ripienaar: That's what this is.  It says its like delete - ie. not a purge - and then expand what is different.

Note Purge has changed, the ADR was Purge() but its not Purge(key) since a bunch of people seemed to think strongly that Purge() isnt needed/wanted due to watches issues.  Purge(key) now is supported as of about 1 hour ago in the server, so this is new behavior, so not a surprise that yours dont match this - it was not possible.

adr/ADR-8.md
scottf: I'm just asking to remove editorialization

ripienaar: No I did not reopen anything. I capture the consensus and added details of HOW I ACTUALLY DID design this.

adr/ADR-8.md
```diff
@@ -92,13 +92,15 @@ type Status interface {
...
 
-	// BucketLocation returns a string indicating where the bucket is located, meaning dependent on backend
-	BucketLocation() string
-
 	// Keys returns a list of all keys in the bucket - not possible now except in caches
 	Keys() ([]string, error)
 
+	// BucketLocation returns a string indicating where the bucket is located, meaning dependent on backend
+	// NOTE: defer implementation until read replica design is finalised
+	BucketLocation() string
```
derekcollison, line 100: I would remove completely for now.

adr/ADR-8.md
```diff
@@ -145,15 +150,21 @@ type KV interface {
 	// Put saves a value into a key
 	Put(key string, val []byte, opts ...PutOption) (seq uint64, err error)
 
-	// Delete purges the subject
+	// Create is a variant of Put that only succeeds when the key does not exist of last historic event is a delete or purge
```
derekcollison, line 153: of -> or

adr/ADR-8.md
```diff
@@ -145,15 +150,21 @@ type KV interface {
...
+	Create(key string, val []byte) (seq uint64, err error)
+
+	// Update is a variant of Put that only succeeds when the most recent operation on a key has the expected sequence
```
derekcollison, line 156: Still prefer revision here and think consistency from here through client impl important.

adr/ADR-8.md
```diff
@@ -180,11 +191,12 @@ A bucket is a Stream with these properties:
  * The main write bucket must be called `KV_<Bucket Name>`
  * The ingest subjects must be `$KV.<Bucket Name>.>`
  * The bucket history is achieved by setting `max_msgs_per_subject` to the desired history level
+ * Safe key purges that deletes history requires rollup to be enabled for the stream using `rollup_hdrs`
  * Write replicas are File backed and can have a varying R value
  * Key TTL is managed using the `max_age` key
  * Duplicate window must be same as `max_age` when `max_age` is less than 2 minutes
  * Maximum value sizes can be capped using `max_msg_size`
- * Maximum number of keys cannot currently be limited - we will add something to the server
+ * Maximum number of keys cannot currently be limited
```
derekcollison, line 199: We should post an issues against server, this one feels like would be used alot.

adr/ADR-8.md
```diff
@@ -236,15 +259,15 @@ subject.  Thus a read for `CONFIGURATION.username` becomes a `io.nats.jetstream.
 ##### History
 
 These operations require access to all values for a key, to achieve this we create an ephemeral consumer on filtered by the subject
-and read the entire value list in using `deliver_all`.
+and read the entire value list in using `deliver_all`. Use a Ordered Consumer to do this efficiently.
```
derekcollison, line 262: Use an?

adr/ADR-8.md
```diff
@@ -254,17 +277,27 @@ place a new message in the subject for the key with a nil body and the header `K
 This preserves history and communicate to watchers, caches and gets that a delete operation should be handled - clear cache,
 return no key error etc.
 
-#### Purging a bucket
+#### Purging a key
+
+Purge is like delete but history is not preserved. This is achieved by publishing a message in the same manner as Delete using the
+`KV-Operation: PURGE` header but adding the header `Nats-Rollup: sub` in addition.
+
+This will instruct the server to place the purge operation message in the stream and in the same RAFT update delete all messages
```
derekcollison, line 285: Would remove RAFT, impl detail. Just say performed as a single server operation or something like that.

derekcollison (approved):
LGTM - Just some minor comments.

adr/ADR-8.md
```diff
@@ -311,18 +323,74 @@ do the right thing with regard to watches and ranges.
...
-We specifically do not place the `Put()` value into local cache since an Entry  must know the JetStream sequence.
+We specifically do not place the `Put()` value into local cache since an Entry must know the JetStream sequence.
 
 This design preserves the read-after-write promises.
 
+#### API Design notes
+
+The API here represents a minimum, languages can add local flavour to the API - for example one can add `PutUint64()` and `GetUint64()`
+in addition to the `Get()` and `Put()` defined here, but it's important that as far as possible all implementations strive to match
+the core API calls - `Get()`, `Put()` and to a lessor extend `Delete()` and `Purge()` and `Entry` - the rest, admin APIs, and supporting
```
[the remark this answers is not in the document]
aricart: and lesser - instead of lessor

PR #66, opened 2021-11-04
[adding] ADR-20 Object Store

scottf:


adr/ADR-19.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+### ObjectStore 
+
+Storing large objects efficiently
+
+```go
+type ObjectStore interface {
+    // Put will place the contents from the reader into a new object.
+    Put(obj *ObjectMeta, reader io.Reader, opts ...ObjectOpt) (*ObjectInfo, error)
```
ripienaar, line 231: I think like KV we should specify the core API and leave convenience functions up to specific languages, so the `GetBytes` and `GetString` variants can be left as a thing where languages do what works for them rather than specify here which implies its needed

adr/ADR-19.md
```diff
@@ -0,0 +1,281 @@
...
+    // Replicas indicates how many storage replicas are kept for the data in the bucket
+    Replicas() int
+    
+    // Sealed indicates the stream is sealed and cannot be modified in any way
+    Sealed() bool
+    
+    // Size is the combined size of all data in the bucket including metadata, in bytes
+    Size() uint64
+    
+    // BackingStore provides details about the underlying storage
```
ripienaar, line 200: For now the only valid value is `JetStream` later we might add or other backing stores.

adr/ADR-19.md
```diff
@@ -0,0 +1,281 @@
...
+where 
+
+`limited-term-w-sp` is defined as 1 or more of `A-Z, a-z, 0-9, dash, underscore, fwd-slash, equals, space`. 
+
+For an individual object name, a dot `.` is allowed except for the first or last character.
+
+The object name (`os-object-name`) is _sanitized_ by replacing any dot or space in the name with an underscore so it
+can be used in the Meta Info message subject 
+
+> It is under consideration to expand object name to allow printable, which would require some kind of encoding when creating the `os-object-name`.
```
ripienaar, line 63: for real world use we'll need this i think, we cant realistically say you cant put `.foo` in a bucket

adr/ADR-19.md
scottf: Ok I'll add notes in the comment to say optional or maybe separate out an optional section.

adr/ADR-19.md
scottf: will add a commment

adr/ADR-19.md
scottf: I was just echoing what was implemented in go, but since we are sanitizing the dot anyway for use in the subject, it doesn't really matter if it's at the beginning or end. I'll change this.

scottf: We are going to change from sanitize to url encoding, which will cover starting or ending with a dot.

ripienaar: Just because its in nats.go does not mean its done :) I'd say we have several open areas for discussion in the engineering group, given that it was designed in a vacuum without input.

ripienaar: So my comment reflects one of the outstanding design issues I think needs addressing before this should be considered for client parity

adr/ADR-6.md
```diff
@@ -28,27 +28,29 @@ A subject is comprised of 1 or more tokens. Tokens are separated by "." and can
...
-dash             = "-"
-underscore       = "_"
-equals           = "="
-printable        = all printable ascii (33 to 126 inclusive)
-term             = (printable except dot, asterisk or gt)+
-limited-term     = (A-Z, a-z, 0-9, dash, underscore, fwd-slash, equals)+
-restricted-term  = (A-Z, a-z, 0-9, dash, underscore)+
-prefix           = (printable except dot, asterisk, gt or dollar)+
-filename-safe    = (printable except dot, asterisk, lt, gt, colon, double-quote, fwd-slash, backslash, pipe, question-mark, ampersand)
+dot                = "."
```
aricart, line 31: We need to go away from all these possible scapes - the name is just an ID, and the ID should just be base64 - client can interpret use this anyway it fits its logic

adr/ADR-19.md
```diff
@@ -0,0 +1,273 @@
...
+Chunk ids (`chunk-id`) should be a nuid.
+
+### Component Templates
+
+| Component | Template |
+| --- | --- |
+| Stream Name | `OBJ_<os-bucket-name>` |
+| Object Info Stream subject | `$O.<os-bucket-name>.M.>` |
+| Chunk Stream subject | `$O.<os-bucket-name>.C.>` |
+| Object Info message subject | `$O.<os-bucket-name>.M.<os-object-name>` |
```
aricart, line 65: not quite sure why we are storing the object name here - effectively if the stream supported multiple files, this would be a really good idea, because then you could list the objects by scanning this subject pattern.

I am thinking that the name of the object should be the bucket name and whether that is a file or not is orthogonal to the functioning of the object store.

ripienaar: 1 bucket can hold many files

adr/ADR-19.md
```diff
@@ -0,0 +1,273 @@
...
+## Naming Specification
+
+Protocol Naming Conventions are fully defined in [ADR-6](ADR-6.md)
+
+### Object Store
+The object store name or bucket name (`os-bucket-name`) will be used to formulate a stream name and is specified as: `restricted-term` or 1 or more of `A-Z, a-z, 0-9, dash, underscore`
+
+### Objects
+An individual object name is custom url encoding to form a portion of a subject (`os-object-name`).
+There are essentially no restrictions on the object name except that it has to be a valid filename across operating systems, so it is specified as:
```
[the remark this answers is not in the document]
scottf: It filename safe so it can be stored on any operating system. It gives a pretty wide latitude of what the name can be. Base64 would work as an encoding, but given the data in a filename that has to be encoded, url encoding only will regularly encode the dot, so it's much shorter.

adr/ADR-19.md
scottf: I agree, `<os-bucket-name>` is redundant after the stream name. They are only in that exact bucket - or are they. What if those messages get moved to a mirror which could be another stream name? Is that right?

scottf: `<os-object-name>` is to allow multiple files per stream (bucket / store)

adr/ADR-19.md
aricart: I am still seeing the variants RI is describing...

aricart:
Also - if we need to do specific things in order to enable these to be easily accessed from different accounts (via imports/exports), we need to include that information here otherwise we'll have a couple of passes on this. @matthiashanel  can perhaps provide some hints.

adr/ADR-19.md
scottf: I had added this "Please note that anything that is commented as a "convenience function" is recommended but optional if it does not make sense for the client language."

adr/ADR-19.md
```diff
@@ -0,0 +1,268 @@
...
+Possible future features
+
+- Event Notifications (add/delete/lock)
+- Locking
+- Archiving (tiered storage)
+- Searching/Indexing (tagging)
+- Versioning / Revisions
+- Overriding digest algorithm 
+- Capturing Content-Type (mime type)
+- Per chunk Content-Encoding (i.e. gzip)
```
jgaskins, line 38: Are y'all considering the ability to get only a given portion of an object? It seems to be a common access pattern for retrieving an object in a lot of blob stores. A lot of object types are fetched in whole (typically medium-sized objects like image files or larger CSVs) and from [my own experimentation](https://jgaskins.dev/nats/NATS/Objects.html) the NATS Object Store works _fantastic_ for that. But some are only fetched in part because fetching them in full would be infeasible. For example:

- browsers fetch video data in ranges to facilitate scrubbing in a `<video>` element via the [HTTP `Range` header](https://httpwg.org/specs/rfc7233.html#range.requests)
- databases can store collections in objects on S3
  - Honeycomb seems to do this, based on talks and blog posts they've published
  - if records are all the same size, it can seek directly to the desired one via offsets

I have a proof of concept using the object store with video and it does work, but the browser makes _many_ `Range` requests for video data, often faster than they can be completed since I'm fetching the entire video from the store.

scottf: Since we ensure we do not delete any messages from the underlying stream, we could provide an API to get a specific stream sequence, so  based on the chunk size you could figure out which slice, but not I don't see supporting a generic range. I considered this for the exact same reason you are, to get a slice of a video which is how all modern video streaming works and I agree that we should support this.

aricart (approved):
LGTM

PR #71, opened 2021-11-18
[adding] ADR-20 API prefixes function and adaptation to materialized JetStream views

matthiashanel:
Signed-off-by: Matthias Hanel <mh@synadia.com>


adr/ADR-20.md
```diff
@@ -0,0 +1,73 @@
...
+Assume the API prefix for JetStream has been set to `JS.from-acc1`.
+For JetStream specific API calls, the local API prefix `$JS.API` has to be replaced with `JS.from-acc1`.
+For the KV API we prefix `$KV` with `JS.from-acc1`, resulting in `JS.from-acc1.$KV`. 
+Thus, in order to put `key` in the bin `bin`, we send to `JS.from-acc1.$KV.bin.key` 
+
+When crossing the account boundaries, this is then translated back to `$KV.bin.key`.
+Thus, the underlying stream still needs to be created with a subscription to `$KV.bin.key`. 
+
+Domains are just a special case of API prefixes and will work the same way.
+The API prefix `$JS.<domain-name>.API` will lead to `$JS.<domain-name>.API.$KV.bin.key`.
```
ripienaar, line 59: But for domains the client libraries would then need to parse the new domain subject as the rewriting wont happen right? Or will you also rewrite there as per account boundary?

adr/ADR-20.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Consequences 
+
+The proposed change is backwards compatible with materialized views already in use.
```
ripienaar, line 63: Generally not super apposed to this, but we have this problem with any stream already in all cases.  I dont really want to fix the problem just for our special streams.

scottf: Yeah, since those ($KV,$OBJ) are just (parts of) stream names, why would we change the name. Why are these different than any other stream and how it is handled within / across a domain? The API calls we make just treat these streams like any other.

ripienaar: These are for the stream listen subject not names - but yeah, that's the general sentiment.

adr/ADR-20.md
matthiashanel: Rewriting happens for domains as well. same as line 33. But I'll add a sentence for clarification.

adr/ADR-20.md
matthiashanel: Which problem are you referring to?

ripienaar: If 2 streams in 2 domains have the same interest subjects both get the message.  This is the problem you are addressing with the `$KV...` subjects.  But users have the same problem with their own streams.

matthiashanel: We are not changing any streams names.
We just change the subjects you use to interact with a resource such that in the current account or domain, they are not received. (assuming they are different)
On entering the target account or entering the target domain, we change them back to what they would have been, had the client been connected and sent them there directly.

What this is about is talking to the right JetStream. Nothing else.
And it is doing so in the very same way we already do.
What I suggest here is the adaptation of an existing technique.

That this also solves the stream matching issue is mere coincidence because the subjects used to subscribe are part of the API.
To me the problem of a message matching possibly multiple streams is a very different one.
One, that frankly only needs addressing for domains.

I do have ideas on how to solve this, but I don't think this ADR is the forum for that.

ripienaar: Yes, please read again, I did mention subjects not stream names. 

Since this is a generic problem for any stream Derek said we should instead find a generic solution.  We should not solve this only for our streams.

matthiashanel: Scott did mention stream names

matthiashanel: While I see this being an issue for domains, cross accounts I don't think there is anything we can do. It's a different account, so would you not store a message in account B, just because it was stored in account A already?
Since this issue does not really apply across accounts, we should support the api prefix, just because it is the current solution. 
What is your opinion on that?

matthiashanel: Thus, supporting the prefix has to happen just to support the cross account use case.

matthiashanel: As for the cross domain use case, in a config based setup, after the push consumer change when cross account stuff is simplified, a user could just have a transit account, from which it only exports API traffic into the actual account used. Isolated like this, all the traffic is fine.

In a jwt based world we'd need the templating mechanism, such that the domain name can be adapted to the domain name of the servers the account is running in. 

If we wanted to push this even further and make it less export/import intensive, I would frankly add a flag to the account named local_scope. When true, all traffic in and out of a leaf node connection where the neighboring domain names differ is denied, except for traffic sent to the domain specific api prefix. 
An application can then choose the setting according to the behavior it requires.
Frankly, this has nothing to do with JetStream either. 
If I have a service foo and set local_scope true, I can reach the service local to main domain under foo and the one offered under the same name in a different domain under say: $EXT.domain.foo

With this, the user can then write domain scoped apps and communicate that to us with a setting. 
We take over the addressing portion externally from their app, with a name that already exists.

matthiashanel: I bet this made you cringe. 
I also would not want to have the resulting hour long discussions about that in the critical path for KV or object. 

Therefore, I would focus on this for now:
We need to support the API prefix in order to deal with the cross account use case.
So the necessary client changes have to happen just because of that.

Domains are a special case of the cross account use case that require a two line server change.

Since one side has to happen anyway and the other side is small and straight forward, I would try to not open a whole other can of worms just yet.

derekcollison: For cross account, we have three main patterns for base JetStream, and hence upper level abstractions like KV. 

1. Placing things into a stream
2. Consuming from a stream via pull
3. Consuming from a stream via push (which is under a design change atm)

All 3 above are (or in case of 3 will be) a service export, meaning the importing account already has control of subject collisions since during import you can remap these.

So, in case of KV, let's say account A has a bucket called foo. So that means the subject is `$KV.foo.>`.
Let's say account B also has a foo KV bucket but wants to access account A. 

During import I could say remap the account A service (not stream because we get a PubAck response), I could ask it to show up as `$KV.foo_A.>`. 

Now I am not sure if we could, inside account B, simple say bind to `foo_A`, but that should work imo.

What I would like to understand better is the following, is the suggested change part of simply making binding to an asset properly leveraging the import remapping we already have, or is this something different altogether?

matthiashanel: @derekcollison It is leveraging what we already have and is following the pattern already established form streams.

derekcollison (approved):
LGTM

scottf (approved):
LGTM

PR #73, opened 2021-11-28
[change] ADR-8 document purge deletes

scottf:


adr/ADR-8.md
```diff
@@ -164,6 +164,9 @@ type KV interface {
 	// Purge removes all data for a key including history, leaving 1 historical entry being the purge
 	Purge(key string) error
 
+	// Completely remove all keys that are in the deleted or purged state, leaving no entries behind for those keys.
+	PurgeDeletes() error
```
ripienaar, line 168: We decided not to add this to the required API. So doesn’t really need to be here or should be marked as optional

scottf: Ok I'll make it optional so we can capture the comment as to what it does

ripienaar (approved):
LGTM

aricart (commented):
Just wondering why we don't just overload `Purge(key)` to do this - ie `Purge("")`

ripienaar:
> Just wondering why we don't just overload `Purge(key)` to do this - ie `Purge("")`

This is purge without tombstones that breaks watchers.  Derek was concerned that if someone deletes/purges 100s or 1000s of keys they end up with 1000s of tombstones.  This is essentially a admin maintenance rather than general use function.

That's why its not in the ADR since its largely there to help the CLI offer the facilty.

PR #76, opened 2021-12-06
[change] ADR-15 ack handling

scottf:


adr/ADR-15.md
```diff
@@ -177,6 +177,20 @@ Why not append those new tokens at the end? This is to simplify the export/impor
...
+
+An ack is accomplished by publishing a message using the JSAck as a subject 
+with the payload consisting of the bytes of the Ack Type
+
+Clients must support 3 terminal Ack Types, `+ACK`, `-NAK` and `+TERM` and the progress Ack Type `+WPI`.
+
+Once a terminal Ack Type has been published for a message, the client should ignore all other 
+user requests to send any type of ack. 
+
+If the consumer is configured with Ack Policy of `none`, the client should ignore all requests to send any type of ack.
```
aricart, line 190: isn't this a noop in the server

kozlovic: Likely, but in the same spirit that after a ack has been published, the library should ignore all other acks, then it would make sense to suppress acks for ack policy of "none". Although in that case, it would actually probably make more sense to return an error, because really the user should know that ack'ing a message does not make sense with ack none: simply ignoring gives a false sense that the user is in control of when the message is considered fully processed (or otherwise redelivered).

aricart (approved):
LGTM - I think the server should ignore any ack/etc sent if the client is set to ack none. Client shouldn't have to add code to try to align or prevent those apis from calling.

kozlovic:
> I think the server should ignore any ack/etc sent if the client is set to ack none

Like I said, it likely does, but that's not the point. At best, you would want the library to avoid sending an ACK that is not needed (reducing network traffic).

But my argument is that in this specific case, I think it may be worth returning an error if user is trying to ACK a message from a subscription that is "ack none" because it shows a misalignment between the intent of the consumer vs what the application is doing: if the user ACKs a message after some processing thinking that until the message is ack'ed, if the app crashes or processing simply fails, the message would be redelivered, then they are going to have a surprise.

But again, even for other ack policy, *user* ack'ing the message twice could also return an error because it means that the user's code is doing something likely wrong.

kozlovic:
After some discussion, my latest vote would be for library to mark a message that has been ack'ed (except for in progress) so that the library ignores the next ACK or auto-ack.

adr/ADR-15.md
```diff
@@ -177,6 +177,20 @@ Why not append those new tokens at the end? This is to simplify the export/impor
...
 
+### Ack Handling
+
+An ack is accomplished by publishing a message using the JSAck as a subject 
+with the payload consisting of the bytes of the Ack Type
+
+Clients must support 3 terminal Ack Types, `+ACK`, `-NAK` and `+TERM` and the progress Ack Type `+WPI`.
+
+Once a terminal Ack Type has been published for a message, the client should ignore all other 
+user requests to send any type of ack.
```
wallyqs, line 188:
```suggestion
Clients should mark that the client has already acked a message, with the exception of the +WPI.  This will help specially with the issue of auto acking and async callbacks, in case NAK, TERM, or ACK is called in the callback, then the client will notice the mark that the client has already attempted to ack and skip sending another ack.  Acking twice using the same message struct may throw an `nats: message was already acknowledged` error, which in case of async callbacks that is doing auto acking would be discarded.
```

Reference to Go implementation: https://github.com/nats-io/nats.go/blob/main/js.go#L2589-L2593

scottf: Do we have to put it in the adr that the client has to track this, I think it's clear that clients will have to track it somehow and I know the C/Java/.NET already do exactly this.
I think we have also agreed for the client to silently ignore repeated calls and not return an error

PR #80, opened 2021-12-14
[adding] ADR-21 NATS Configuration Contexts

ripienaar:

### DIFF
--- a/adr/ADR-21.md
+++ b/adr/ADR-21.md
@@ -1,0 +1,119 @@
+# NATS Configuration Contexts
+
+|Metadata|Value|
+|--------|-----|
+|Date    |2021-12-14|
+|Author  |@ripienaar|
+|Status  |Partially Implemented|
+|Tags    |client|
+
+## Background
+
+A `nats context` is a named configuration stored in a configuration file allowing a set of related configuration items to be stored and accessed later.
+
+In the `nats` CLI this is used extensively, for example `nats stream ls --context orders` would load the `orders` context and configure items such as login credentials, servers, domains, API prefixes and more.
+
+The intention of the ADR is to document the storage of these contexts so that clients can, optionally, support using them.
+
+## Version History
+
+|Date|Revision|
+|----|--------|
+|2020-08-12|Initial basic design|
+|2020-05-07|JetStream Domains|
+|2021-12-13|Custom Inbox Prefix|
+
+This reflects a current implementation in use widely via the CLI as such it's a stable release.  Only non breaking additions will be considered.
+
+## Design
+
+Today the design is entirely file based for maximum portability, later we can consider other options like S3 buckets, KV stores etc.
+
+### Configuration Paths
+
+There is generally no standard for what goes on in a users home directory on a Unix system. Recently the Free Desktop team have been working on [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) that specifies in detail where configuration, data, binaries and more are to be stored in a way that's compatible with Linux desktops like KDE and Gnome but also with systems such as systemd.
+
+We therefore based the design on this specification as a widely supported standard.
+
+|File|Description|
+|----|-----------|
+|`~/.config`|The default location for user configuration, configurable using `XDG_CONFIG_HOME`|
+|`~/.config/nats`|Where all NATS related user configuration should go|
+|`~/.config/nats/context.txt`|The current selected (default) context, would contain just `ngs`|
+|`~/.config/nats/context/ngs.json`|The configuration for the `ngs` context|
+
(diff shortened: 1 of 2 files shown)


ColinSullivan1 (commented):
Let's discuss Q1 22. It's a great feature, we just need to determine where it'll be implemented.

ripienaar:
> Let's discuss Q1 22. It's a great feature, we just need to determine where it'll be implemented.

👍 for now this is just capturing the feature, not communicating an intend for clients to implement.

aricart (approved):
LGTM

aricart (approved):
LGTM

adr/ADR-21.md
```diff
@@ -0,0 +1,119 @@
...
+|`ca`| |Path to the x509 Certificate Authority|
+|`nsc`| |A `nsc` resolve url for loading credentials and server urls|
+|`jetstream_domain`| |The JetStream Domain to use|
+|`jetstream_api_prefix`| |The JetStream API Prefix to use|
+|`jetstream_event_prefix`| |The JetStream Event Prefix|
+|`inbox_prefix`| |A prefix to use when generating inboxes|
+
+All fields are optional, none are marked as `omitempty`, users wishing to edit these with an editor should known all valid key names.
+
+Above settings map quite obviously to client features with the exception of `nsc`, the `nsc` key takes a URL like value, examples are:
```
matthiashanel, line 71: Should that nsc be invoked with store and nkey directory specified by the context as well?

ripienaar: Quite happy to let those who choose to customise everything carry that  complexity burden themselves by just pointing at the credential file themselves. 

This infinite customizability we are adding at every level is a plague. 

Why would nsc need a context? @aricart and I discussed that long ago and decided it’s not compatible models

ripienaar: Also, afaik, nsc reads environment variables and a config file, not options you pass? `NSC_HOME` and `NKEYS_PATH` etc, I am not going to be supporting passin these around. If someone sets them they will continue to work for nsc when invoked by the context.

tbeets:
In sample usage API, the value of OS environment variable would be the filename of the desired context or the friendly name of the desired context?   Both options would seem useful.   @ripienaar

ripienaar:
> In sample usage API, the value of OS environment variable would be the filename of the desired context or the friendly name of the desired context? Both options would seem useful. @ripienaar

No file name based access - its just the context name. In future we'd have non file based context storage, so its only name.

tbeets:
LGTM.  I think whether we factor in more NSC stuff entirely related to futures of CLI tools (i.e. what and how many).

derekcollison:
I think this direction and exposing this to apps and scripts and automation is spot on.

One observation I have had is that a bunch of popular tools just own `~/.<proj>`, so for us would be `~/.nats` maybe.

ripienaar:
> One observation I have had is that a bunch of popular tools just own `~/.<proj>`, so for us would be `~/.nats` maybe.

That's the nice thing about standards, we can just follow the it without thinking what some others tools or dont. The standard covers everything from local data storage, binaries and more - storage paths for example that we need for things like `nats server run` and integrates with the eco system at large.

Why would we choose something that specifically would work against the desktop os direction?

derekcollison:
My $0.02 is that default standards outweigh the formalized ones, who continually seem to be losing power. So wanted to see what the popular projects with lots of users were doing. I will check if anyone uses that convention today or if they have noted they would switch, but I am not aware of any. Are you?

ripienaar:
It has VERY WIDE adoption. Few kind of small projects like Emacs, curl, git, GTK, helm, ImageMagick, less and a pile of standard shell things, every linux desktop environment, systemd and its dynamically mounted home directories, mercurial, mutt, nano, neovim, python pip, tmux, etc.

derekcollison:
Good to know. Been using emacs forever so it is still `~/.emacs` for me. A bunch of the other ones as well, but maybe new installs default to ``~/.config/<project>`

ripienaar:
Yes, obviously backwards compat on tool like the listed ones is fairly important, but the gravity is towards this. Emacs checks `~/.config/emacs` in adition to older paths.

tbeets:
I see about 50/50 sticking their stuff under ~/.config by default.   JetBrains, Google Chrome, and NATS CLI are three examples that do.   Rust, Mozilla, Maven, minikube, AWS, and Azure are examples of those that splat in the homedir.    Lots of variability.

tbeets:
Definitely support sticking application-specific stuff under `~/.config` to separate from typical shell config etc. in homedir if we have to make a default location call either way.

ripienaar:
Arch wiki has a partial list of things that supports it https://wiki.archlinux.org/title/XDG_Base_Directory#Supported

jnmoyne (approved):
LLANBTM

(Looks like a no brainer to me)

PR #82, opened 2021-12-15
adding api prefix note

matthiashanel:


ripienaar (approved):
LGTM

adr/ADR-8.md
```diff
@@ -37,6 +37,7 @@ additional behaviors will come during the 1.x cycle.
  * Custom Stream Names and Stream ingest subjects to cater for different domains, mirrors and imports
  * Key starting with `_kv` is reserved for internal use
  * CLI tool to manage the system as part of `nats`, compatible with client implementations
+ * Accept arbitrary application prefixes, as outlined in [ADR-19](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-19.md)
```
ripienaar, line 40: To clarify here @matthiashanel we say arbitrary prefixes, not yet the domain behaviors right? I dont want to support magical domain behaviors now I think there's significant enough concern about the current design in ADR-19 about that aspect.  So for now just arbitrary prefixes?

matthiashanel: @ripienaar A few days ago, the two of us agreed on that. Focus on the cross account use case and ignoring domains for now.

Yes, the prefixes are arbitrary.

ripienaar: Ok yeah just confirming. I wasn’t sure if we agreed to domain behaviours or not. Thanks!

matthiashanel: np, enjoy the holidays

PR #88, opened 2022-01-24
ADR-22: Client file locations

philpennock:
In discussion, it became clear that folks are more comfortable using a common directory holding things if there's some way to avoid multiple tools claiming the same files.  There was also debate about how to manage migration safely, how to manage Windows, etc.  I volunteered to write an ADR documenting a proposal.

At heart, we're using XDG on Unix with the <https://github.com/adrg/xdg> rules for mapping these locations to "native" locations on macOS and Windows.

This ADR has a proposal as a stake in the ground which folks can evaluate.

(Also, fixed the ADR template and parser to be a little more consistent, after I went wrong while following the template).

### DIFF
--- a/adr/ADR-22.md
+++ b/adr/ADR-22.md
@@ -1,0 +1,197 @@
+# Client File Locations
+
+|Metadata|Value|
+|--------|-----|
+|Date    |2022-01-24|
+|Author  |@philpennock|
+|Status  |Approved|
+|Tags    |client, config|
+
+
+## Context and Problem Statement
+
+The configuration files for NATS clients should be in a coherent predictable location.
+Clients in various languages should be able to know where to read and write data,
+and administrators should know what file-system access is needed and where
+sensitive key data might be stored.
+
+This document both unifies the location and acts as a living registry of the
+namespace within the NATS configuration directory.
+
+
+## Version History
+
+|Date      |Revision|
+|----------|--------|
+|2022-01-24|Initial proposal|
+
+
+## [Context | References | Prior Work]
+
+* [ADR-21][] on Configuration Contexts moves us closer to the model
+  described herein, but ADR-21 will need updating for the non-XDG platforms.
+* The `nsc` and `ngs` commands use other locations, scattering files around.
+* A Unix-centric model is the [XDG Base Directory Specification][xdgbase].
+
+
+## Design
+
+We mandate a standard schema for where all NATS clients should lay out their
+configuration files, and guidance for migration and handling.
+
+Our goals are:
+
+ 1. Keep files of like type together, while adhering to OS norms.
+ 2. Never break existing user configuration.
+ 3. Provide a predictable working path across Unix-like systems, both macOS
+    and XDG-compliant Unix system.
+ 4. Register which files and sub-directories exist, to act as a central
+    repository of such information, and record who can safely rely upon such
+    details.
+
+Rather than reinvent the wheel, we propose to use the layout schema of
+[Adrian-George Boston's Go implementation of XDG][adrg/xdg], which uses
+specific other locations for other operating systems, all documented in
+tables in the front README.
+
+The documentation is specific enough to allow implementations in languages
+other than Go to use the same layouts.  For Go, the `adrg/xdg` implementation
+has two dependencies, both of which we already depend upon in multiple
+`nats-io` repositories: `github.com/stretchr/testify` and `golang.org/x/sys`.
+The license is MIT and the code is stable but still seeing maintenance
+updates.  It supports, “most flavors of Unix, Windows, macOS, and Plan 9”.
+
+All configuration should live inside a directory called `nats`.  Users should
+not need to learn which other tools might drop files elsewhere and should be
+able to, minimally, discover all the configuration by files or symbolic links
+within such directories.  For configuration files, on Unix, this means that
+all per-tool configuration will be within the `~/.config/nats/` directory,
+while managed data might be within the `~/.local/share/nats/` directory.
+Environment variables and choice of OS will change that.
+
+Our one deviation from the [adrg/xdg][] pattern is that on macOS, we
+additionally require that `~/.config/nats` work, but in new installs it should
+be a symbolic link to the `~/Library/Application Support/nats` directory.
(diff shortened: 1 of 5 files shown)


adr/ADR-22.md
```diff
@@ -0,0 +1,184 @@
...
+Much of the existing client configuration involves authentication data such as
+NKeys or combined `.creds` files.  These are more likely than most to have
+been moved by the users to secure storage and replaced with symbolic links or
+be on encrypted mounts.  As such, automatically migrating such data is fraught
+with peril.
+
+For the time being, we instead simply mandate that files be accessible by the
+new paths; it is acceptable for client libraries to detect that the new
+location does not exist on disk but the old location does exist, and put in a
+symbolic link at the new location pointing to the old location.
```
ripienaar, line 82: Granted I have like 10 apps using `~/.config` but I dont see any that make such symlinks - and I know I use several that support XDG but ignore it if old files are found.  Is there existing example of such symlinks (not against them, just curious)

ripienaar (approved):
LGTM

adr/ADR-22.md
philpennock: I think the difference is in whether or not the app considers the location something which has to be shared with other apps.  If it's an internal detail, then just using the old location is fine.  But if we're setting expectations for where all clients, across implementation languages, should look, then ensuring that "the required location" has a symlink to "the old location" removes the need for every client library implementation to know about historical variants to just work.

main.go
```diff
@@ -28,7 +28,7 @@ type ADR struct {
 }
 
 var (
-	validStatus = []string{"Approved", "Partially Implemented", "Implemented"}
+	validStatus = []string{"Approved", "Partially Implemented", "Implemented", "Proposed"}
```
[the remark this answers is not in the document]
philpennock: Ah.  Then since the README says to run this, I guess the fix is to change the README and the template text, then revert the Go change?

adr/ADR-22.md
```diff
@@ -0,0 +1,184 @@
...
+|Date      |Revision|
+|----------|--------|
+|2022-01-24|Initial proposal|
+
+
+## [Context | References | Prior Work]
+
+* [ADR-21][] on Configuration Contexts moves us closer to the model
+  described herein, but ADR-21 will need updating for the non-XDG platforms.
+* The `nsc` and `ngs` commands use other locations, scattering files around.
```
aricart, line 33: `ngs` tool doesn't write any state file (pretty sure)
`nats` cli is the only other cli tool

aricart:
it would be great if @derekcollison could comment on this. As I understand as I understand his comments (he can elaborate) most unix tools don't use these locations. On my environment the only clis that I can find in OS X that store under `~/Library/Application Support` are `termshark` and `mkcert` all other tools have an UI type component with the exception of the demon/services in OS X.

main.go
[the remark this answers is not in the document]
ripienaar: No doubt the README can be improbed you're not the first that made this error - happy to see edits there to improve it

philpennock:
I have reverted the `main.go` changes and adjusted the README and template, per @ripienaar's feedback above.

In a meeting with @derekcollison, @aricart and @scottf we settled on the approach documented now: we use the adrg/xdg locations, because particularly on Windows the XDG locations do not work.  For darwin/macOS we compromise and want both the XDG location and the system-native paths to work.  Tools should mostly use the system native path, the .config/nats works better for humans exploring.  We think.

Suggestions on better phrasing welcome.

adr/ADR-22.md
```diff
@@ -0,0 +1,197 @@
...
+    - Most applications should avoid assumptions.
+    - Contexts can contain sensitive data, where a server uses password-based
+      authentication.
+    - compatibility location: strict XDG adherence on all platforms.
+  + The `nsc/` and `nkeys/` directories are reserved for client credentials
+    and the account system, as described above.
+    - Most tools should not need to be specifically aware in code of the
+      location of these; where credentials are passed in, they are typically
+      part of the site configuration.
+    - New tools should use [ADR-21][] configuration contexts to locate
```
ripienaar, line 177: fyi this repo will change `ADR-21` to a link, so you can take away the markdown reference etc

adr/ADR-22.md
aricart: @philpennock can you clarify:

> additionally require that `~/.config/nats` work, but in new installs it should
be a symbolic link to the `~/Library/Application Support/nats` directory.

But what are we doing for windows? Are we doing to then honor what the library says? Note that this is awesome, but it is somewhat fastidious, as we LGTM'ed to use the library which does the correct thing for in any OS (the addition of the symlink is fine I think) but the trick is it was rejected - I have since completed, and released, and this will do additional tweaks - and in the case of windows will put things completely elsewhere.

adr/ADR-22.md
```diff
@@ -0,0 +1,197 @@
...
+
+The configuration used by the `nats` CLI are the configuration contexts, which
+are generic for client tooling and not specific to that tool, so we consider
+the `context/` directory and `context.txt` file to be top-level configuration
+items.
+
+### The `nsc` CLI
+
+The default locations of files for `nsc` can be overridden with the
+environment variables: `$NSC_HOME`, `$NKEYS_PATH`, and `$NSC_CWD_ONLY`.
```
aricart, line 124: $NSC_CWD_ONLY shouldn't be part of this - since that enables a feature that allows multiple terminal instances to have the right context in terms of the /Operator/Account model. This is an app feature, not some standard.

adr/ADR-22.md
```diff
@@ -0,0 +1,197 @@
...
+
+In addition, `nsc` has traditionally detected an account system Operator in
+the current directory and then ignored all environment variables to use the
+current directory as a root.
+This behavior is considered an authentication administrator mode and should
```
aricart, line 129: Correct this is user - the reality is the data locations on a local user's directory is not the intended idea - The JWT configurations managed by nsc are probably shared between multiple groups - ie - should be in github.

aricart (approved):
LGTM - but generally want some discussion that is explicit for windows.

PR #92, opened 2022-02-10
Add GetRevision()

kozlovic:
GetRevision returns a specific revision value for the key, or "KeyNotFound" if that message for this revision can't be found. See https://github.com/nats-io/nats.go/pull/903

### DIFF
--- a/adr/ADR-8.md
+++ b/adr/ADR-8.md
@@ -143,7 +143,7 @@ The interface here is a guide of what should function in read-only mode.
 type RoKV interface {
 	// Get gets a key from the store
 	Get(key string) (Entry, error)
-
+	
 	// History retrieves historic values for a key
 	History(ctx context.Context, key string) ([]Entry, error)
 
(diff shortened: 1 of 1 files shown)


kozlovic:
@ripienaar Not sure if this is needed in the ADR. Regardless, since we have added that to the Go client, I will create an issue so that other libs can add it.

ripienaar:
I think to make it for the ADR 

> @ripienaar Not sure if this is needed in the ADR. Regardless, since we have added that to the Go client, I will create an issue so that other libs can add it.

I think to be suitable for the ADR we should consider rounding out the history feature set a bit - for example a API call to take revision x and promote it to the current value for a key.

I kept it out initially as I wanted to see how the history / audit feature shakes out.  Could be fine to mark this as optional also and leave as is?

kozlovic:
>  Could be fine to mark this as optional also and leave as is?

Just searched for "optional" in that whole PR and don't find it, so that would be the only one with that comment, so not sure this is the right approach then. Again, I have created the issue so that libraries can provide this feature, but maybe it is indeed optional and does not have to be in the ADR. In other words, happy to close this PR without a merge..

kozlovic:
@ripienaar I removed from the kv-readonly section and added as a note where we say that we can add GetUint64() or the like. Feels a bit lost there and makes me really wonder if it is worth updating the ADR.
Again, since this is considered optional, then I am not certain we need to update this, simply have the issue that I have created so that devs can implement it if they want their lib to have the API..

ripienaar:
Yeah I agree, if its optional I wouldnt bother with the ADR. Later when we choose to add a set of audit/history related management things it makes sense in the ADR, in isolation its a bit low value

kozlovic:
Closing then, since there is no value adding it there. Thanks!

PR #98, opened 2022-02-18
update config / minor change

scottf:


matthiashanel (approved):
I assume this is based on your testing last year?
LGTM

scottf:
> I assume this is based on your testing last year? LGTM

This config is what @aricart is using in his test. The key difference is the prep for inbox prefix

PR #105, opened 2022-03-19
Add notes on implementation of js.Publish retries

wallyqs:
From base implementation at  https://github.com/nats-io/nats.go/pull/930

Signed-off-by: Waldemar Quevedo <wally@synadia.com>

### DIFF
--- a/adr/ADR-22.md
+++ b/adr/ADR-22.md
@@ -85,16 +85,24 @@ if err != nil {
 }
 ```
 
-### Make Publish retry as needed until context deadline
+### Make Publish retry as needed until deadline
 
-In Go when using the context package, it can be possible to set the maximum deadline of the retries
-so that the client can retry as needed.  In the example below a client will attempt to publish for 10 seconds
-backing off `250ms` as needed until the service is available again:
+It can be possible to set the maximum deadline of the retries so that the client can retry as needed.
+In the example below a client will attempt to publish up to 10 seconds to wait for an ack response
+from the server, backing off `250ms` as needed until the service is available again:
 
 ```go
+// Using Go context package
 ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 defer cancel()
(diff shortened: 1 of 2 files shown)


adr/ADR-22.md
```diff
@@ -0,0 +1,101 @@
...
+|----------|---------------------------|
+| Date     | 2022-03-18                |
+| Author   | wallyqs                   |
+| Status   | Partially Implemented     |
+| Tags     | jetstream, client         |
+
+## Motivation
+
+When the NATS Server is running with JetStream on cluster mode, there
+can be occasional blips in leadership due which can result in a number
```
kozlovic, line 13: typo: `leadership due which`

adr/ADR-22.md
```diff
@@ -0,0 +1,101 @@
...
+### Make Publish retry as needed until context deadline
+
+In Go when using the context package, it can be possible to set the maximum deadline of the retries
+so that the client can retry as needed.  In the example below a client will attempt to publish for 10 seconds
+backing off `250ms` as needed until the service is available again:
+
+```go
+ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+defer cancel()
+_, err := js.Publish("foo", []byte("bar"), nats.Context(ctx), nats.RetryWait(250*time.Millisecond), nats.RetryAttempts(-1))
```
kozlovic, line 97: Does that mean that this is applicable only to "context" feature?

What I mean by that is that without a context, the Publish() is still subject to the js "timeout" or "wait" timeout (that all internal APIs requests are subject to, using it as the timeout of the Request() call). So the question is: should libraries without a context still apply the same logic but use the request timeout as the upper limit?

kozlovic (changes requested):

adr/ADR-22.md
wallyqs, line 13:
```suggestion
can be occasional blips in leadership which can result in a number
```

kozlovic (approved):
LGTM

PR #126, opened 2022-07-03
ADR-24 Restricted Inbox

tbeets:
PR created for visibility on [ADR-24](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-24.md) in-process branch. [ADR-24](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-24.md) defines a new feature to create a server-enforced restricted inbox that only a specific connected client can subscribe to. The PR should be approved only when [ADR-24](https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-24.md) specification is complete.

### DIFF
--- a/adr/ADR-24.md
+++ b/adr/ADR-24.md
@@ -1,0 +1,17 @@
+# Restricted Inbox  
+
+|Metadata| Value                     |
+|--------|---------------------------|
+|Date    | 2022-06-29                |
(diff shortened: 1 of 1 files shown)


ripienaar:
Doesn't appear this got anywhere, will close, though still desirable.

PR #133, opened 2022-07-18
ADR-30 Subject Transform

tbeets:
Provide architectural coverage of a server implementation that is integral to several user-level features in the server.

### DIFF
--- a/adr/ADR-30.md
+++ b/adr/ADR-30.md
@@ -4,7 +4,7 @@
 |----------|-------------------------|
 | Date     | 2022-07-17              |
 | Author   | @derekcollison, @tbeets |
-| Status   | `Implemented`           |
+| Status   | Implemented             |
 | Tags     | server                  |
 
 ## Context and Problem Statement
--- a/adr/ADR-30.md
+++ b/adr/ADR-30.md
@@ -2,7 +2,7 @@
 
 | Metadata | Value                   |
 |----------|-------------------------|
-| Date     | 2022-17-07              |
+| Date     | 2022-07-17              |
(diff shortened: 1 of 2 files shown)


adr/ADR-30.md
```diff
@@ -0,0 +1,61 @@
...
+
+For a valid input subject:
+
+* Source-matching literal-token positions are ignored, i.e. do not appear in the output subject
+* Source-matching wildcard-token positions (if any) are _placed_ in the output subject in positions defined by the Destination format
+* Source-matching `*` wildcard (single token) tokens must be placed in Destination format by wildcard cardinal position number using `$x` or `{{wildcard(x)}}` notation
+* Source-matching `>` wildcard (multi token) tokens are mapped to the respective `>` position in the Destination format
+* Literal tokens in the Destination format are mapped to the output subject unchanged (position and value)
+
+### Example transforms
```
ripienaar, line 46: nats server mapping command might be worth a mention.

tbeets: Do you mean something different/added apart from mention of "Subject mapping and partitioning" higher up?

ripienaar: Sorry should have been clearer the NATS CLI has “nats server mapping” that lets you test and interact with these and see how they would behave. 

I don’t recall now if it’s in the current shipped version - next for sure and JNM did great work in surfacing better errros for it

tbeets: Very cool!  That is so handy!  I will make a reference.

derekcollison (approved):
LGTM

PR #141, opened 2022-08-08
ADR-31 Direct Get

tbeets:
This is an implementation ADR for the Direct Get and Subject-Appended Direct Get features.

### DIFF
--- a/adr/ADR-31.md
+++ b/adr/ADR-31.md
@@ -99,14 +99,15 @@ Direct Get may return an error code:
 - `404` if the request is valid but no matching message found in stream 
 - `408` if the request is empty or invalid
 
+> Error code is returned as a header, e.g. `NATS/1.0 408 Bad Request`. Success returned as `NATS/1.0` with no code.
+
 Otherwise, Direct Get reply contains the message along with the following message headers:
 - `Nats-Stream`: stream name
 - `Nats-Sequence`: message sequence number 
 - `Nats-Time-Stamp`: message publish timestamp
 - `Nats-Subject`: message subject
-- `Nats-Last-Sequence`: sequence number of previous message for the given subject (if any)
 
-A _regular_ (not JSON-encoded) NATS message is returned (from the stream store).
+> A _regular_ (not JSON-encoded) NATS message is returned (from the stream store).
 
 ## Example calls
 
--- a/adr/ADR-31.md
+++ b/adr/ADR-31.md
@@ -1,0 +1,182 @@
+# JetStream Direct Get 
+
+| Metadata | Value                                         |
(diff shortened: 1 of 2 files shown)


adr/ADR-31.md
```diff
@@ -0,0 +1,182 @@
...
+
+When Allow Direct is true, each of the stream's servers will also subscribe to `$JS.API.DIRECT.GET.<stream>.>` with
+fixed queue group `_sys_`.  Requests to this API endpoint will be interpreted as a shortcut for `last_by_subj` request
+where `last_by_subj` is derived by the token (or series of tokens) following the stream name rather than request payload.
+
+It is an error (408) if a client calls Subject-Appended Direct Get and includes a request payload.
+
+#### Response Format 
+
+Direct Get may return an error code:
```
kozlovic, line 98: Just maybe add that it is through the use of headers, not a typical JetStream response object.

adr/ADR-31.md
```diff
@@ -0,0 +1,182 @@
...
+- `404` if the request is valid but no matching message found in stream 
+- `408` if the request is empty or invalid
+
+Otherwise, Direct Get reply contains the message along with the following message headers:
+- `Nats-Stream`: stream name
+- `Nats-Sequence`: message sequence number 
+- `Nats-Time-Stamp`: message publish timestamp
+- `Nats-Subject`: message subject
+- `Nats-Last-Sequence`: sequence number of previous message for the given subject (if any)
```
kozlovic, line 107: I don't think that this applies to direct get (more for republish feature I think?)

kozlovic (changes requested):
Would remove reference to `Nats-Last-Sequence` since this does not apply to direct get (it is for republish) and maybe add a comment that the possible error returned is through a nats header, not a API error type of thing.

adr/ADR-31.md
aricart: The direct headers are these:
```
export enum DirectMsgHeaders {
// the stream where the message is stored
  Stream = "Nats-Stream",
// the sequence number of the stored message
  Sequence = "Nats-Sequence",
// the timestamp of the stored message
  TimeStamp = "Nats-Time-Stamp",
// the subject that was ingested
  Subject = "Nats-Subject",
}
```

tbeets: Thanks.  Will remove.  Copy/paste from Issue 115 that did seem kind of awkward for this API.

adr/ADR-31.md
tbeets: Yes, I show in an example but good suggestion to clarify here.  Will add.

PR #145, opened 2022-08-12
Draft logical permissions

bruth:


adr/ADR-32.md
```diff
@@ -0,0 +1,179 @@
...
+  APP: {
+    users: [
+      {
+        user: operator,
+        password: operator,
+        permissions: {
+          allow: [
+            "inbox(operator)",
+            "js-stream-operator(EVENTS)",
+          ]
```
wallyqs, line 129: Maybe also need to cover how this maps to the imports and export syntax while we are at it, sub represent either a `stream` or `service`.
Like being able to define that reading info about consumer and streams is ok from the other account but not ok to create, something like: 

```hcl
  exports: [{stream: "foo"}, { jetstream: { streams: [ { name: EVENTS, export: [stream-info, consumer-info] }]
```

adr/ADR-32.md
```diff
@@ -0,0 +1,179 @@
...
+Fundamentally, this design is a light abstraction on top of the current way permissions are defined in configuration or JWTs. These new constructs can be converted to a standard permissions map internally.
+
+There are two potential layers to defining user permissions. First, a logical name must be defined for the underlying API subject. This mapping is shown in the table below. In addition to the name, the _context_ (publish or subscribe) can also be inferred. For example, all of the JS APIs use request-reply initiated by the client. So `js-create-stream` is implicitly a publish-based permission. For subjects that are *parameterized* based on stream names, consumer names, or subjects, the `(...)` syntax is used to support specifying these parameters.
+
+Category | Name    | Subject | Pub/Sub | Notes
+---------|---------| ------- | ------- | -----
+General  | `pub(subject)` | `{subject}` | pub | Potential alternative to a bare subject string.
+\* | `sub(subject)` | `{subject}` | sub | Same as above.
+\* | `inbox` | `_INBOX.>` | sub
+\* | `inbox(id)` | `_INBOX_{id}.>` | sub | *Convention for custom inboxes, so be used with `--inbox-prefix` for private access.*
```
wallyqs, line 39: need to be able to express the permissions for ACKing responses from the consumer, maybe there could be a macro for js-consumer to be able to subscribe and ack

adr/ADR-32.md
```diff
@@ -0,0 +1,179 @@
...
+        }
+      },
+      {
+        user: joe,
+        password: joe,
+        permissions: {
+          allow: [
+            "inbox(joe)",
+            "pub(services.*)",
+          ]
```
wallyqs, line 151: this syntax is nice, I was thinking of granular permission like below but seems very verbose:

```hcl
    permissions = {
      # Can publish to stream
      publish: { allow = ["a"] } 

      # Granual permissions for JetStream features
      jetstream: {
        streams { EVENTS = { allow: [ stream-info ], deny: [add-stream, delete-stream] } }
        kv { TEST = { allow: [ put, del], deny: [ destroy-kv ] }
        object-store { allow: [...] }
      }
    }
   }
```

bruth: I like the structure that your example provides which provides namespacing (as I am doing `js-` and `kv-`). A future state I considered was user-defined permissions which may break the structure. For example, if a new key called `roles` within an account is defined, you could declare new mappings off the primitive logical permissions the NATS server provides.

adr/ADR-32.md
bruth: [These](https://github.com/nats-io/nats-architecture-and-design/pull/145/files#diff-c13b1651f7f36a6775ce53f919367c98427da78d1b669f27a90eae0423beab2aR92-R95) correct?

adr/ADR-32.md
bruth: Indeed. Fundamentally, these permissions can be thought of as functions that expand to one or more subjects `js-stream-info(EVENTS)` &rarr; `$JS.API.STREAM.INFO.EVENTS`. I thought about having these permissions able to be intermingled with standard subjects. If that is the case, the ambiguity of subject vs perm needs to be removed, so some kind of prefix character could help differentiate, such as `@` or whatever.

Then you could have something like:

```
exports: [
  {service: "@js-stream-info(EVENTS)"},
]
```

The implicit pub/sub information would really only apply in the user permissions context.

adr/ADR-32.md
wallyqs: ah yes those...

adr/ADR-32.md
aricart: It should be possible to subset keys/object names on kv/object store - currently object store doesn't expose data chunks in a way that can be clamped.

PR #146, opened 2022-08-22
Object Store refinement

scottf:


scottf:
Need to add Placement and RePublish

adr/ADR-20.md
```diff
@@ -38,60 +38,98 @@ Possible future features
...
 
-### Chunk Ids
-Chunk ids (`chunk-id`) should be a nuid.
+### Object Name
+An individual object name is not restricted. It is base64 encoded to form `name-encoded`.
 
-### Component Templates
+### Digest
+Currently `SHA-256` is the only supported digest. Please use the uppercase form as in [RFC-6234](https://www.rfc-editor.org/rfc/rfc6234)
+when specifying the digest as in `SHA-256=3F9239B0272558CE6E3E98731C43A654AC6882C5050D5F352B2A5BFBB8DE0058`.
```
aricart, line 65: Note the hash may have padding `=` or `==` or `===`

aricart (dismissed):
split into 2 different prs please - we have KV and ObjectStore

adr/ADR-20.md
```diff
@@ -216,54 +290,73 @@ is recommended but optional if it does not make sense for the client language.
 ```go
 type ObjectStore interface {
     // Put will place the contents from the reader into a new object.
-    Put(obj *ObjectMeta, reader io.Reader, opts ...ObjectOpt) (*ObjectInfo, error)
+    // It is an error to Put when ObjectMeta contains a Link. Use AddLink or AddBucketLink
```
aricart, line 293: Not sure why we are copying the go interface here. Any functionality that beyond put/get that returns byte is specific to the client and can be optional implementation.

Instead, put the basic interface for the object store, put, get, link, delete, watch, seal, status, and then put a link to the go code.

scottf: I have addressed these comments.

adr/ADR-20.md
scottf: fixed example

aricart (approved):
LGTM - I would have described the models and interactions not the Go types.

PR #159, opened 2022-09-23
Update ADR 8 KV Status interface

jnmoyne:
Adds `Bytes()` to the bucket Status interface (see https://github.com/nats-io/nats.go/pull/1092)


tbeets (approved):
LGTM

PR #165, opened 2022-09-30
ADR-4 Proposal: alternate character encodings in headers

caleblloyd:
NATS users may want to send header keys and values other in encodings other than ASCII.  This is especially popular in languages other than English.

Highlight that header key/value encodings that are backward-compatible with ASCII can optionally be offered by NATS Clients.

Allow clients to select UTF-8 as their default encoding, and annotate that UTF-8 is the preferred character encoding.

Understandably, client implementations may rely on HTTP header parsing libraries that enforce ASCII.  For this reason, and also for backward compatibility reasons, ASCII should remain a valid default encoding option.

### DIFF
--- a/adr/ADR-4.md
+++ b/adr/ADR-4.md
@@ -52,6 +52,9 @@ Any case sensitivity in header interpretation is the responsibility of the appli
 
 > Note: This is _different_ from HTTP headers which declare/define that web server and user-agent participants should ignore case.
 
+###### Character Encoding
+NATS Clients _may_ offer alternate header key and value encodings that are backward-compatible with ASCII.  NATS Clients _should_ use a default header key and value encoding of either UTF-8 (preferred) or ASCII.
+
 With above caveats, please refer to the
 [specification](https://tools.ietf.org/html/rfc7230#section-3.2) for information
 on how to encode/decode HTTP headers.


kozlovic (commented):
I would vote against since if you allow 1 client to support UTF-8, then it means that all clients must otherwise you take the risk of inter-operability issues between clients. Also note that for headers that are used by the server, the server is doing a search of given header names without even "unpacking" headers. But this may not be an issue since those headers are already defined and supporting UTF-8 would not change that, but just wanted to point that out.

caleblloyd:
As it turns out, `nats.go` already partially supports UTF-8, at least in the Header Values:

image

scottf:
So are we talking about the client encoding as it puts the key/value into the headers (and then across the wire).
For instance if the user presents `z Ḁ Ḃ Ḉ x` as the actual value, we actually store/transfer it as `\u007a\u0020\u1e00\u0020\u1e02\u0020\u1e08\u0020\u0078` and that somewhere in our headers handling, the user can tell us to do this encoding?

caleblloyd:
@scottf correct, this is to determine how to Encode Header key/value `strings` -> `byte[]` on the wire when sending a message, and Decode Header key/value `byte[]` -> `string` off the wire when parsing a message

scottf:
So that's encoded to ascii, similar to json/url encoding. Java built in string conversion to bytes allows for an encoding such as UTF-8, but for instance these are the actual 15 bytes `122 32 -31 -72 -128 32 -31 -72 -126 32 -31 -72 -120 32 120` it makes for that string, so it's really just double(?) byte, so just making sure that is _not_ what we want.

caleblloyd:
UTF-8 is variable length encoding, between 1-4 bytes per character.  So the []byte count may not be equal to the character count, the header payload length would need to be the encoded bytes length.

ColinSullivan1:
Am not opposed to providing additional encodings but imo the default should be ASCII, which all clients would need to support.  This would allow NATS to better facilitate direct interop with HTTP, gRPC, etc and is the lowest common denominator across languages, tooling, browsers, etc.  Could expand to support RFC 2047 (MIME) as in ISO-8859-1 (https://www.w3schools.com/charsets/ref_html_8859.asp). Is there a specific use case driving the proposal?

More on this here:  https://www.jmix.io/blog/utf-8-in-http-headers/

caleblloyd:
The reason for wanting UTF-8 to be the preferred default is it would allow for much broader character support out-of-the-box in headers while maintaining byte-for-byte compatibility with ASCII if only ASCII characters are in-use.

`string` in most programming languages supports Unicode.  It seems like a very limiting practice to allow the user to create a header collection of `Dictionary<string, string[]>` with full Unicode support, but then limit that down to just 128 ASCII characters.

If folks want to interop with HTTP that is fine, just don't use any non-ASCII characters, and UTF-8 will encode the to the exact same thing as ASCII.

scottf:
+ 1 to provide for optional encoding, but I think default should be restricted as currently defined, not even UTF-8. The caveat being that if we provide a way to optional encode in one client, this really becomes a parity item that all clients must provide optional encoding.

aricart:
The reason for the keys to be ASCII is because that is the HTTP and MIME header specification.
one question on this is whether this is customer driven or we just want to be able to support it. The main issue as Scott points out, the minute that a client writes one of these encoded things, client parity for all clients becomes an issue.

Jarema:
I would think that this is pretty rare use case, especially that NATS headers are perceived as being pretty close to HTTP headers.

Not being actually HTTP header spec compatible already caused problems in quite a few languages that wanted to reuse really well optimized and tested `http headers` libraries, but couldn't (as our NATS headers for JS/KV/ObjectStore are required to be sent case-sensitive to the server).
This change could make things harder.

I see one more reason: as we're having some ideas around nice NATS microservices tooling, connectors and more, straying further from HTTP headers spec could bring even more problems.

I think the simplest solution is to just suggest users to keep Keys according to HTTP headers spec, which should not be a big issue, and if they need non-asci values, just base64 them.

caleblloyd:
I don't think it is that rare of a use case, say an app wants to pass around `X-User-Name` and someone's name is in a different language not covered by ASCII.

But upon further reading of the spec, it seems that optional encoding for field values.  Excerpt from [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values) for Field Values:

> Field values are usually constrained to the range of US-ASCII characters [[USASCII](https://www.rfc-editor.org/rfc/rfc9110.html#USASCII)]. Fields needing a greater range of characters can use an encoding, such as the one defined in [[RFC8187](https://www.rfc-editor.org/rfc/rfc9110.html#RFC8187)].

So maybe this isn't even needed then, it is already laid out in the spec.  ASCII should probably be the default, but clients "can use an encoding"

Issue #49, opened 2021-08-17
JS.ACK new tokens


kozlovic:
## Overview

Newer servers will include possibly 2 tokens to the `$JS.ACK.` ack subject, a possible domain, and an always present "account hash". The library does not do anything with those, but need to take them into account in order to properly locate the tokens that it needs.

The behavior is documented in ADR-15, section: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-15.md#js-ack

## Clients and Tools

 - [x] Go @wallyqs
 - [x] Java @scottf
 - [x] JavaScript 
 - [x] .Net @ColinSullivan1 @scottf
 - [x] C @kozlovic
 - [x] Python @wallyqs
 - [x] Rust @spacejam
   
## Other Tasks

 - [x] docs.nats.io updated
 - [x] Update ADR to `Implemented`
 - [x] [Update client features spreadsheet](https://docs.google.com/spreadsheets/d/1VcYcKqwOp8h8zZwNSRXMS5wrdA1jZz6AumMTHZbXrmY/edit?usp=sharing)

Client authors please update with your progress. If you open issues in your own repositories as a result of this request, please link them to this one by pasting the issue URL in a comment or main issue description.

bruth:
Opened nats-io/nats.py#351 for Python support.

bruth:
Docs PR https://github.com/nats-io/nats.docs/pull/485

marthaCP:
Colin updated ADR 15; All actions are complete as of 10/27/22

PR #168, opened 2022-10-27
Mark ADR-15 as Implemented

ColinSullivan1:
Per https://github.com/nats-io/nats-architecture-and-design/issues/49, it looks like we can mark this as implemented and close out the issue.

Signed-off-by: Colin Sullivan <colin@synadia.com>

### DIFF
--- a/adr/ADR-15.md
+++ b/adr/ADR-15.md
@@ -4,7 +4,7 @@
 |--------|-----|
 |Date    |2021-08-11|
 |Author  |@kozlovic|
-|Status  |Partially Implemented|
+|Status  | Implemented|
 |Tags    |jetstream, client|
 
 ## Context


bruth:
Is there a separate ADR for the server implementation for this? If I recall correctly, @wallyqs had noted it was [not yet implemented on the server](https://github.com/nats-io/nats-server/blob/ad0a8403bb286369c612267b9907af2580fcfa40/server/jetstream_api.go#L222).

ripienaar:
closing this, we should instead mark all this as deprecated in favor of https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-37.md I'll see about adding a new field to adr statusses

ripienaar:
Deprecated status and marking this ADR as deprecated in https://github.com/nats-io/nats-architecture-and-design/pull/293

PR #176, opened 2022-11-23
js simplification ADR

aricart:
This is the formalization of the JetStream simplification goals into the ADR. Some details still missing on the options but drawing the line in the sand.


adr/ADR-32.md
```diff
@@ -0,0 +1,117 @@
...
+low level `ConsumerInfo`.
+
+This means that in some languages `add` may not be available (unless major
+version bump in the library). For that reason libraries can choose how to
+instantiate a consumer:
+
+- `new Consumer(consumerInfo)`
+- `add(streamName, consumerOptions)`
+- `create(streamName, consumerOptions)`
+- `get(streamName, consumerName)`
```
scottf, line 55: what about delete? It is already in management, but would it be nice to just have near these api

adr/ADR-32.md
```diff
@@ -0,0 +1,117 @@
...
+
+By their very nature, OrderedConsumers don’t play well with the above API for
+retrieving the consumer, because the consumer itself will be created and
+possibly re-created by the client during it's lifecycle. Clients may wish to
+provide a:
+
+`getOrderedConsumer(streamName, startOptions, bufferingOptions)` which allows
+consuming messages while conforming to the API operations on a consumer outlined
+below.
+
```
scottf, line 73: Does this mean we need to prevent ordered consumers in the new/add/create?
Also, maybe it's better we present Consumer as _ConsumerContext_ 
1) In Java, the name Consumer is already taken, so I'm using ConsumerContext anyway
2) It decouples the actual consumer/info from Consumer/ConsumerContext so ordered is just a variety of Consumer/Context. Also leaves the door open for other things we have not thought of yet.

scottf:
It would be nice to add in the part about the Consume payload not just being a message, but an object/tuple whatever

adr/ADR-32.md
aricart: sure

adr/ADR-32.md
aricart: you can create them, but the object that you return must be able to be recreated at a different sequence.

aricart:
> It would be nice to add in the part about the Consume payload not just being a message, but an object/tuple whatever

This is language specific.

scottf:
> > It would be nice to add in the part about the Consume payload not just being a message, but an object/tuple whatever
> 
> This is language specific.

Agreed the implementation is language specific. I'd just like to call it out specifically in the doc that the behavior is required.

Jarema (approved):
LGTM!

piotrpio (approved):
LGTM!

aricart:
@Jarema @piotrpio @scottf added an initial discussion of the options - note that there are some edge conditions which will need to address.

adr/ADR-37.md
```diff
@@ -27,37 +27,75 @@ performance.
...
-implementation, typically by providing a stream name and some consumer options.
-
-Clients wanting to consume messages from the consumer, simply locate the
-`Consumer`, which then provides functionality for processing, or consuming
-messages. The representation of the `Consumer` is retrieved, and provides
-operations to `fetch` and `consume` messages.
+The JetStream API consists of three main components: JetStreamContext, Streams,
+and Consumers.
+
+### JetStream Context
```
scottf, line 33: This stuff used to live in JetStream Management Context. Seems unnecessary to change that.

adr/ADR-37.md
```diff
@@ -27,37 +27,75 @@ performance.
...
+- operations used to consume messages:
+  - `consume()`
+  - `fetch()`
+  - `next()`
+
+### Design and naming in individual client libraries
+
+The design explained above should be implemented in client libraries with
+respect to each language's and libraries bes practices. Method names may differ
+(e.g. add vs create), as well as palcement of some of the methods (e.g. in some
```
scottf, line 93: palcement -> placement

adr/ADR-37.md
```diff
@@ -27,37 +27,75 @@ performance.
...
+  - `getStream(streamName)`
+  - `deleteStream(streamName)`
+  - `listStreams()`
+- operations on consumers:
+  - `addConsumer(streamName, consumerConfig)`
+  - `getConsumer(streamName, consumerName)`
+  - `deleteConsumer(streamName, consumerName)`
+- `accountInfo()`
+
+### Streams
```
scottf, line 54: Does anyone feel like this just ends up hiding the server meaning of stream. Developers are already used the the word context. A context makes sense to me since it represents a stream and operations that can be done with and around stream.

PR #183, opened 2022-12-28
add response type hints to micro responses

ripienaar:
This adds type hints in line with those used by JetStream API to micro responses.

The goal is to be able to attach JSON schema (and others) to these for validation, documentation, discovery of design and more.

These schemas will be surfaced in "nats schema" command as well.

### DIFF
--- a/adr/ADR-32.md
+++ b/adr/ADR-32.md
@@ -95,6 +95,10 @@ All discovery and status responses contain the following fields:
 
 ```typescript
     /**
+     * An identifier of the message type for example io.nats.micro.v1.stats
+     */
+    type: string,
+    /**
     * The kind of the service reporting the status
     */
     name: string,
@@ -115,6 +119,7 @@ Returns a JSON having the following structure:
 
 ```typescript
 {
+    type: string,
     name: string,
     id: string,
     version: string,
@@ -137,12 +142,15 @@ All the fields above map 1-1 to the metadata provided when the service was
 created. Note that `subject` is the subject that the service is listening for
 requests on.
 
+The type for this is `io.nats.micro.v1.info_response`.
+
 ### PING
 
 Returns the following schema (the standard response fields)
 
 ```typescript
 {
+    type: string,
     name: string,
     id: string,
     version: string,
@@ -152,6 +160,8 @@ Returns the following schema (the standard response fields)
 The intention of `PING` is for clients to calculate RTT to a service and
 discover services.
 
+The type for this is `io.nats.micro.v1.ping_response`.
+
 ### SCHEMA
 
 Returns a JSON having the following structure. Note that the `schema` struct is
@@ -159,6 +169,7 @@ only returned if the `schema` was specified when created.
 
 ```typescript
 {
+    type: string,
     name: string,
     id: string,
     version: string,
@@ -177,6 +188,8 @@ only returned if the `schema` was specified when created.
     };
 }
 ```
+
+The type for this is `io.nats.micro.v1.schema_response`.
 
 ### STATS
 
@@ -215,6 +228,8 @@ only returned if the `schema` was specified when created.
     started: string
 }
 ```
+
+The type for this is `io.nats.micro.v1.stats_response`.
 
 ## Error Handling
 


adr/ADR-32.md
```diff
@@ -216,6 +229,8 @@ only returned if the `schema` was specified when created.
 }
 ```
 
+The type for this is `io.nats.micro.v1.stats_response`.
```
aricart, line 232: Since the stats can be modified - the stats handler can attach additional data, should this be customizable by the stats handler?

aricart (approved):
LGTM

adr/ADR-32.md
ripienaar: Maybe though the `data` is free form so if the handler wants to mark it up its welcome to mark it up, in the schema I'll just say its anything?

adr/ADR-32.md
```diff
@@ -94,6 +94,10 @@ Services should respond to:
 All discovery and status responses contain the following fields:
 
 ```typescript
+    /**
+     * An identifier of the message type for example io.nats.micro.v1.stats
+     */
+    type: string,
```
scottf, line 100: @aricart @ripienaar Why do we need this, we already have name, which is described as "kind". We don't need both, but maybe we just change `name` to `type`

scottf (changes requested):

adr/ADR-32.md
ripienaar: It's to identify some random piece of JSON that you might not exactly know where its from (you could for example publish these stats message into a stream regularly or to some remote service). 
It's for documentation. 
It's to help others wanting to write clients.  
It's to communicate design decisions.  
It's consistent with jetstream

unversioned json is just amateur hour

adr/ADR-32.md
scottf: @ripienaar What do you mean by "free form" I have already suggested that data just be a string, That string could be whatever, it could happen to be json, it could happen to be the word "true", it could happen to be a number. Just like our message payload, just like our kv value. 
As it is now, we had to build a complicated way to inject a data handler to handle a json anything, whereas if it is just string, all I have to worry about is make sure I json encode the string properly since it's a string in a json object.

scottf (changes requested):

adr/ADR-32.md
ripienaar:
```
returns unknown data that can be serialized as JSON
```

it's anything.  So I wont mark up the specific type of the included data, the type here only conveys that its a micro status message.  What random stuff people put in the extra data is their business

scottf: The doc is not clea or complete, `data` is not "anything", it is json, meaning it is expected that the right side of the `"data":` is an actual json object, array, number or boolean, string etc., but not "just" a string.
This is why it requires a handler to encode and decode and this is why it's a huge pain. Not all languages have built in support for json, which is what this requires because it means passing a generic language specific json object to the user (or in the case of java/.net I have to send a string of the contents to the encoder/decoder which is double processing) otherwise requiring a callback to handle. 
It would be easier / less complex to code and to understand if it `data`'s type was string, instead of being fancy.

ripienaar: The spec is clear, something that can be serialized as JSON, it implies its valid JSON.

I have no idea why you are discussing the data field on this PR its unrelated. I am adding a type hint for the Stats object.

adr/ADR-32.md
scottf: We already have a generic data field in stats. Also, again, why would one instance of a service "foo" have a `type`, where another instance of "foo" have a different `type`. 
Recently it was requested that we add a `meta` data field to the info. Is this the same purpose? 
Also, why does the `type` need to be of every response ping/info/stats/schema

ripienaar: Ideally the service backend can publish advisories:

 * At start: info
 * At stop: stats
 * Regularly: stats

Very commonly these kinds of advisories are picked up and routed to other services where things like what NATS subject it was obtained from is gone.  Then passed to things like azure/aws event gateways which route messages based on data from their content.

As it stands, there is no way to know what a message is if you are given just the message.

adr/ADR-32.md
scottf: I bring it up because the idea of a generic `type` is parallel to `data`, and if a generic string is good enough for `type` why isn't it good enough for `data`

scottf: > Maybe though the `data` is free form so if the handler wants to mark it up its welcome to mark it up, in the schema I'll just say its anything?

ripienaar: The type describes this:

```go
	Stats struct {
		ServiceIdentity
		NumRequests           int             `json:"num_requests"`
		NumErrors             int             `json:"num_errors"`
		LastError             string          `json:"last_error"`
		ProcessingTime        time.Duration   `json:"processing_time"`
		AverageProcessingTime time.Duration   `json:"average_processing_time"`
		Started               string          `json:"started"`
		Data                  json.RawMessage `json:"data,omitempty"`
	}
```

There's a LOT to describe and derive etc and much to use in addition to the `data` and the kind describes this.

scottf: That's from the go client impl. It's nice and convenient for go to have that built in type. Java and .NET do not have an equivalent or even a json interface to lean on, so those clients have quite a bit of complicated work to simulate that.

adr/ADR-32.md
aricart: > Why do we need this, we already have name, which is described as "kind". We don't need both, but maybe we just change `name` to `type`

one is an API type, the other is marking the data for processing by something that doesn't know what that response is. This give the possibility of other tooling like nats-cli to do something smart with that response

adr/ADR-32.md
ripienaar: `json.RawMessage` isnt anything special, its just `[]byte` that holds valid JSON.  If Java can validate something is valid JSON you can a) get from stats handler a string b) validate its JSON c) store it as `string` in the data.

scottf: Java and .NET cannot natively validate something is json.

PR #185, opened 2022-12-30
ensure time is in the same tz for global deployments

ripienaar:


aricart (approved):
LGTM - for JavaScript all ISO is always UTC, but other languages may not do it.

PR #186, opened 2023-01-05
Update Service API ADR with multi-endpoints details

piotrpio:


adr/ADR-32.md
```diff
@@ -33,8 +34,8 @@ Service configuration relies on the following:
 - `statsHandler` - an optional function that returns unknown data that can be
   serialized as JSON. The handler will be provided the endpoint for which it is
   building a `EndpointStats`
-- `endpoint` - a subject and a handler (effectively equivalent to a NATS
-  subscription)
+- `endpoints` - a collection of endpoints (name + handler) to be associated with the
+  service. Those can be either passed in configuration, or addded via a method on service.
```
scottf, line 38: Can we just simplify this and only have it as part of the configuration of the service and not as something later? Are their use cases where they will make a root and then sometimes later change to endpoint? Or If they make endpoints, will they add another later?
What about instead of the ability to add and endpoint later (like when the service is already running), we just have the ability to turn on and off an already configured/running endpoint?

adr/ADR-32.md
```diff
@@ -132,9 +137,9 @@ Returns a JSON having the following structure:
      */
     version: string,
     /**
-     * Subject where the service can be invoked
+     * Root subject of the service
      */
-    subject: string
+    root_subject: string
```
scottf, line 142: Would it be useful to have another field for a list of endpoints, which may be empty?

adr/ADR-32.md
```diff
@@ -223,10 +242,6 @@ The type for this is `io.nats.micro.v1.schema_response`.
     * Average processing_time is the total processing_time divided by the num_requests
     */
     average_processing_time: Nanos;
-    /**
-    * ISO Date string when the service started
-    */
-    started: string
```
scottf, line 229: just verifying that this is to be removed

adr/ADR-32.md
```diff
@@ -270,7 +285,17 @@ Note the name of the queue group is fixed to `q` and cannot be changed otherwise
 different implementations on different queue groups will respond to the same
 request.
 
-The handler specified by the client to process requests should operate as any
+For each configured endpoint, a queue subscription should be created. Subscription
+subject is composed by concatenating root subject and endpoint name:
+
+```typescript
+{$root_subject}.{$endpoint_name}
+```
```
scottf, line 293: Should we be validating the root_subject and endpoint_name the same as service name (`A-Z, a-z, 0-9, dash, underscore`)? Or maybe a different validation, but still a validation?

scottf (changes requested):

adr/ADR-32.md
piotrpio: I was thinking about this and if we want to configure endpoints together with the service, it may make more sense to have `started` on the top level (for the whole service, not particular endpoints).

adr/ADR-32.md
piotrpio: Hmm, possibly we should validate, but I think not the same way as service name, but rather the same way as any NATS subject - we may have subject delimiters or wildcards in `root_subject`.

adr/ADR-32.md
piotrpio: Not sure I get it - `endpoints` is this field. Or you mean something else?

adr/ADR-32.md
piotrpio: From what I know, some languages which do not use callbacks for services, but rather iterators (e.g. Rust) do not have the option to easily store endpoints in config, so there it is easier to have it as a method on service interface. It is language specific I guess, for Go I'm leaving endpoint setup on `Config`.

adr/ADR-32.md
scottf: another field
```
root_subject: string
endpoints: list of string
```

piotrpio: Ah, sorry, I thought this comment was on config... Correct, I think having endpoints on INFO is a good idea.

adr/ADR-32.md
```diff
@@ -132,9 +137,9 @@ Returns a JSON having the following structure:
      */
     version: string,
     /**
-     * Subject where the service can be invoked
+     * Root subject of the service
      */
-    subject: string
+    root_subject: string
 }
```
aricart, line 143: We should have a the root subject, and a list of all the other subjects (as augmented by additional endpoints.

adr/ADR-32.md
aricart: That is what I did already on my PR - started is a common info.

adr/ADR-32.md
aricart: the subscription will fail if they do it wrong, so no validation here.

aricart: Note that it should be possible to do:

`accounts.*` for a root subject, and have endpoints that are:
`add`
`delete`


In that case the wildcard in the root subject is related to an account ID. - The info endpoint can take a number of other arguments. This is still a valid service. Also the endpoint could have a wildcard as well.

adr/ADR-32.md
```diff
@@ -33,8 +34,8 @@ Service configuration relies on the following:
 - `statsHandler` - an optional function that returns unknown data that can be
   serialized as JSON. The handler will be provided the endpoint for which it is
   building a `EndpointStats`
-- `endpoint` - a subject and a handler (effectively equivalent to a NATS
```
aricart, line 36: This part I don't like - I know we had discussed this between @ripienaar and @Jarema. But here's the issue:

- if you want multiple endpoints you cannot have a handler on the root subject (why not - I hear it is possibly not the right thing, etc but this is an assumption - the app developer will figure out what they want)
- if you want a single endpoint you have add a root subject, and then the named handler, the composite is the subject.

This second thing is what I am more opposed to - the easy case should be here's the subject and the handler.

aricart (changes requested):

aricart:
Possibly the compromise is config for the handler can take a `{root_subject: string, endpoints?: NamedEndpoints[]}` or a `{subject: string, handler: ()=>{}}` If you use the first, on some clients now it becomes possible to do `addEndpoint()`, if the second config is used, `addEndpoint()` is an illegal operation - the only problem is developer will have to refactor logic and subjects to support multiple endpoints if they started with the single endpoint.

adr/ADR-32.md
ripienaar: Yeah agree on this, I mentioned earlier I think a single endpoint on the root should be the only one that can specify wildcard subjects and the intention is that it is for just a basic one feature thing (like auth service example I made) or for those making their own request MUX.

If you go multiple entpoints, then you need to follow the convention.

So in some conversation over the holiday period we suggested keeping `endpoint` for the easy path and adding `endpoints` that follow convention - but of course pick another implementation detail that you wish but I agree with @aricart here

aricart:
To summarize the changes:

- `root_subject` is optional. If set and you add endpoints, they hang of the root_subject, otherwise, they are the full subject.
- `addGroup(token: string)` returns an object that you can `addEndpoint()` so you can build your subject hierarchy, so you can do:
```
const srv = addService({root_subject: "service"})...
srv.addEndpoint("stop", stopHandler);
srv.addGroup("v1").addGroup("accounts").addEndpoint("list", listHandler);
srv.addGroup("v1").addGroup("accounts").addEndpoint("delete", deleteHandler):
// or:
const v1 = srv.addGroup("v2");
v2.addEndpoint()..
```
Note `addGroup()` doesn't have handlers or anything, it is just a builder context for the subject hierarchy
- `info()` must return a list of all subjects for user endpoints.
- `stats()` for a particular endpoint must the full concrete subject the endpoint is listening on.

scottf:
What about just getting rid of root subject? Your example already allows an endpoint without a group.
What do you think about "path" instead of "group"?

aricart:
@scottf I think group is better term than path here. Also the thing that you are pointing out is that if you have group, the need for root subject is a moot point. Because since a group can have a name or not, it will act accordingly.

The other benefit to not having that configuration is that the config could now have an endpoint just like it did before and behave correctly - albeit it now becomes optional.

scottf:
1. Maybe all endpoints need a unique name  - I just learned that we are expected to allow multiple endpoints of the same subject, so we need something unique. It could be an id I suppose that we assign and the dev can see.

2. Do we need the whole group thing? Why can't we just 

```
const srv = addService(name:"name")...
srv.addEndpoint("blahName", "blahSubject", blahHandler, <optional> blahStatsDataHandler);
srv.addEndpoint("stop", "control.stop", stopHandler, <optional> stopStatsDataHandler);
srv.addEndpoint("list account", "v1.accounts.list", listHandler);
srv.addEndpoint("delete account", "v1.accounts.delete", deleteHandler);
```

adr/ADR-32.md
aricart: @ripienaar in retrospect after updating the client and writing some tests, the need for endpoint is not necessarily there, the difference and would simplify the control mechanism - previously it would be added right on the `nc.services.add()`, but a separate call to srv.addEndpoint() achieves the same thing.

```typescript
  const srv = await nc.services.add({
    name: "example",
    version: "0.0.1"
  });
  srv.addEndpoint("add", ()=>{});
  const v1 = srv.addGroup("v1");
  v1.addEndpoint("add", () => {});

  const v2 = srv.addGroup("v2");
  v2.addEndpoint("add", () => {});
  v2.addGroup("internal").addEndpoint(("add"), () => {});
```
Note however perverse that looks it is fairly "legal".

The stats look like this:
@piotrpio  note that the endpoint stats _cannot_ be a map, they have to be a list, as the names for the endpoints could be the same but under different subject. Also note that the subject for each endpoint is included.

```javascript
{
  type: "io.nats.micro.v1.stats_response",
  name: "example",
  id: "NC9LTC35N9Q11ZSZE8BA4Z",
  version: "0.0.1",
  started: "2023-01-06T01:19:19.823Z",
  num_requests: 0,
  num_errors: 0,
  processing_time: 0,
  average_processing_time: 0,
  last_error: undefined,
  endpoints: [
    {
      name: "add",
      subject: "add",
      num_requests: 0,
      num_errors: 0,
      processing_time: 0,
      average_processing_time: 0
    },
    {
      name: "add",
      subject: "v1.add",
      num_requests: 0,
      num_errors: 0,
      processing_time: 0,
      average_processing_time: 0
    },
    {
      name: "add",
      subject: "v2.add",
      num_requests: 0,
      num_errors: 0,
      processing_time: 0,
      average_processing_time: 0
    },
    {
      name: "add",
      subject: "v2.internal.add",
      num_requests: 0,
      num_errors: 0,
      processing_time: 0,
      average_processing_time: 0
    }
  ]
}

```

aricart: Note above, the metric type information doesn't exist at the top level of the stats, as these are now tracked per endpoint, we could aggregate, but not sure that is important as the raw data is there, and some of the fields like errors couldn't show last_error etc.

aricart:
> 1. Maybe all endpoints need a unique name  - I just learned that we are expected to allow multiple endpoints of the same subject, so we need something unique. It could be an id I suppose that we assign and the dev can see.
> 2. Do we need the whole group thing? Why can't we just
> 
> ```
> const srv = addService(name:"name")...
> srv.addEndpoint("blahName", "blahSubject", blahHandler, <optional> blahStatsDataHandler);
> srv.addEndpoint("stop", "control.stop", stopHandler, <optional> stopStatsDataHandler);
> srv.addEndpoint("list account", "v1.accounts.list", listHandler);
> srv.addEndpoint("delete account", "v1.accounts.delete", deleteHandler);
> ```

not sure that we need it to be honest - the stats will report the name they gave and the subject calculated - so if these differ, they will be alongside. 

Now to the point of a stats handler, if you do multiple addEndpoint() with the same name or subject, the stats handler won't be able to differentiate between them, but possibly it doesn't need to. If you want to know how many requests went to hello world and you have that sort of topology, then providing the aggregate doesn't seem wrong.

adr/ADR-32.md
aricart: rootSubject won't be part of the config.

adr/ADR-32.md
```diff
@@ -108,9 +109,10 @@ All discovery and status responses contain the following fields:
     id: string,
     /**
     * The version of the service
+    * Should be validated using official semver regexp: 
+    * https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
     */
     version: string
```
aricart, line 115: for the version checking - take the string remove any leading and trailing letters
try to parse digit.digit.digit if it works, it is semver.

adr/ADR-32.md
```diff
@@ -223,15 +246,44 @@ The type for this is `io.nats.micro.v1.schema_response`.
...
 
+## Adding groups and endpoints
+
+A service can be extended by adding additional groups and endpoints.
+
+### Groups
+
+A group serves as a common prefix to all endpoints registered in it.
+A group can be created using `addGroup(name)` method on a Service.
+Group name should be a valid NATS subject, but cannot contain `>`
```
aricart, line 262: _but cannot end in `>`_

adr/ADR-32.md
```diff
@@ -223,15 +246,44 @@ The type for this is `io.nats.micro.v1.schema_response`.
...
+wildcard (as group name serves as subject prefix).
+
```
aricart, line 264: Group can also have an empty `""` name, reason for this is the process for the service can inject a prefix if necessary. Groups that are empty ignore the group name when building the subject. This allows for exporting a service to a particular account, etc.

adr/ADR-32.md
```diff
@@ -270,7 +322,12 @@ Note the name of the queue group is fixed to `q` and cannot be changed otherwise
 different implementations on different queue groups will respond to the same
 request.
 
-The handler specified by the client to process requests should operate as any
+For each configured endpoint, a queue subscription should be created.
+
+> Note: Handler subject does not contain the `$SRV` prefix. This prefix
```
aricart, line 327: Not sure why this note is here, perhaps we should clarify that $SRV is a reserved subject group used for monitoring and other generated functionality.

aricart (approved):
LGTM except for the stated clarifications

adr/ADR-32.md
piotrpio: Well, I suppose it is pretty much the same thing right? I mean, nats subject can only have `>` at the end, group cannot have it anywhere.

adr/ADR-32.md
piotrpio: Yeah, that was the point, just to make sure we do not add the `$SRV` prefix to endpoint handlers. I added this because I saw this done in other PR so just wanted to clarify.

PR #188, opened 2023-01-11
Update Service API service and endpoint schema definitions

piotrpio:
This PR contains changes to how Service API handles schema definitions. 
Instead of one global `Schema` on `Config`, now there are:
- A service-scoped API specification url (passed in config)
- Endpoint-scoped schemas defined when adding each endpoint

### DIFF
--- a/adr/ADR-32.md
+++ b/adr/ADR-32.md
@@ -25,15 +25,11 @@ Service configuration relies on the following:
   `A-Z, a-z, 0-9, dash, underscore`.
 - `version` - a SemVer string - impl should validate that this is SemVer
 - `description` - a human-readable description about the service (optional)
-- `schema`: (optional)
-  - `request` - a string/url describing the format of the request payload can be
-    JSON schema etc.
-  - `response` - a string/url describing the format of the response payload can
-    be JSON schema etc.
+- `apiUrl`: (optional) - url pointing to user-defined API specification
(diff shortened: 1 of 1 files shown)


aricart (approved):
LGTM

ripienaar:
ime its nearly impossible to extract per action schemas out thats actually useful, so I would use the url most often, though might have been nice to be able to return a service specific open api schema that covers the whole thing without needing a url.

though it would also be nice if a url could be like `kv://registry/service.v1` to get it out of kv

piotrpio:
@ripienaar We were thinking about aggregating per-endpoint schemas to form a complete schema, but also give possibility to add a service-scoped schema URL (though I agree, url to KV would be nice).
The way it is proposed now would enable to easily parse all endpoint schemas into some target format, because when you call `SCHEMA`, you get all you want - endpoint names, subjects and request/response schemas.

ripienaar:
Yeah what’s there now seems fine. 

Personally I would just use the top schema feature since as I am sure you know breaking even the most basic OpenAPI or Proto spec into somethings that describes a single function that then results in a valid OpenAPI or proto spec is extremely difficult given the general quality of these libs

But no harm in supporting oer endpoint specs for those who can use tbem as long as the top level one is fully capable

Jarema (approved):
LGTM

We should think about KV support in near future.

PR #189, opened 2023-01-12
Service API base endpoint name

piotrpio:
Add information on the name of base endpoint - if base endpoint subject is provided by the user, its name should be set to "default" by client libraries.


aricart (approved):
LGTM

PR #193, opened 2023-01-19
Ported Derek's callout document as an ADR

aricart:


adr/ADR-26.md
```diff
@@ -0,0 +1,236 @@
+# NATS Authorization Callouts
+
+| Metadata | Value          |
+| -------- |----------------|
+| Date     | 2022-11-28     |
+| Author   | @derekcollison |
+| Status   | `Implemented`  |
```
ripienaar, line 7: no quotes around these - though its such a common mistake i should probably allow it hehe

ripienaar (approved):
LGTM

adr/ADR-26.md
aricart: easier yet, remove the quotes from the template... :)

PR #202, opened 2023-02-14
[adr-##] connect / reconnect logic

scottf:


ripienaar:
Replaced with https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-40.md

PR #204, opened 2023-03-08
ADR-36 Stream subject transformations

jnmoyne:


adr/ADR-XX.md
```diff
@@ -0,0 +1,76 @@
...
+|Metadata| Value                     |
+|--------|---------------------------|
+|Date    | 2023-02-10                |
+|Author  | @jnmoyne                  |
+|Status  | `Partially Implemented`   |
+|Tags    | jetstream, client, server |
+
+## Context and Problem Statement
+
+Subject mapping and transformation is only available at the Core NATS level, meaning that in order to define or modify mappings one has to either have access to the server config file, or have access to the account's key in operator security mode. While Core NATS subject mapping has its place and use (e.g. scaling a single stream for writes using partitioning), many (most) use cases for subject mapping happen in the context of streams, and having to go to the Core NATS server/account level to define subject mappings is quite limiting as it's not easy for an application programmer to be able to define the mappings he/she needs (even if they have access to the account's key).
```
tbeets, line 12: Have we officially defined **Core NATS** and institutionalized?

jnmoyne: Yes as far as I know, there are certainly a number of references to "Core NATS" in the docs (to distinguish with JS)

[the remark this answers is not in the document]
jnmoyne: Ok, added those possible use cases

adr/ADR-XX.md
```diff
@@ -0,0 +1,76 @@
...
+|Metadata| Value                     |
+|--------|---------------------------|
+|Date    | 2023-02-10                |
+|Author  | @jnmoyne                  |
+|Status  | `Partially Implemented`   |
+|Tags    | jetstream, client, server |
+
+## Context and Problem Statement
+
+Subject mapping and transformation is only available at the Core NATS level, meaning that in order to define or modify mappings one has to either have access to the server config file, or have access to the account's key in operator security mode. While Core NATS subject mapping has its place and use (e.g. scaling a single stream for writes using partitioning, traffic routing, A/B and Canary testing), many (most) use cases for subject mapping happen in the context of streams, and having to go to the Core NATS server/account level to define subject mappings is quite limiting as it's not easy for an application programmer to be able to define the mappings he/she needs (even if they have access to the account's key).
```
tbeets, line 12: I think you can strike this paragraph entirely, or just shorten to "Subject mapping and transformation at the time of client publish (PUB) is useful for many use cases; but there are use cases in stream sourcing (or chaining) where transformation features can also be valuable."

adr/ADR-XX.md
```diff
@@ -0,0 +1,76 @@
...
+## Prior Work
+
+See [ADR-30](ADR-30.md) for Core NATS subject mapping and a description of the available subject transform functions.
+
+## Features introduced
+
+The new features introduced by the [PR](https://github.com/nats-io/nats-server/pull/3814) allow the application of subject mapping transformations in two places in the space configuration:
+
+- You can apply a subject mapping transformation as part of a Stream source.
+  - This enables the ability to do sourcing between KV bucket  (as the name of the bucket is part of the subject name in the KV bucket streams, and therefore has to be transformed during the sourcing as the name of the sourcing bucket is different from the name(s) of the bucket(s) being sourced).
```
tbeets, line 25: Would add "Amongst other use cases, ... KV bucket ..."  to avoid this sounding like a KV added feature and ADR.

adr/ADR-XX.md
```diff
@@ -0,0 +1,76 @@
...
+- You can apply a subject mapping transformation at the ingres (input) of the stream, meaning after it's been received on Core NATS, or mirrored or sourced from another stream, and before limits are applied (and it gets persisted). This subject mapping transformation is only that, it does not filter messages, it only transforms the subjects of the messages matching the subject mapping source.
+  - This enables the ability to insert a partition number as a token in the message subjects.
+
+![](images/stream-transform.png)
+
+In addition, it is now possible to source from the same stream more than once.
```
tbeets, line 31: Nice to have an example use case, e.g. if upstream has "foo", "bar", and "baz" but the downstream only wants "foo" and "baz" than this can be achieved by two source definitions (one for "foo" and one for "baz").

adr/ADR-XX.md
```diff
@@ -0,0 +1,76 @@
...
+    "dest": "mapped.foo.>"
+  },
+  "sealed": false,
+  "deny_delete": false,
+  "deny_purge": false,
+  "allow_rollup_hdrs": false,
+  "allow_direct": false,
+  "mirror_direct": false
+}
+```
```
tbeets, line 71: Would be good to have an example with multiple source definitions, with perhaps being two from the same upstream but with different filter_subject.

adr/ADR-XX.md
jnmoyne: I would rather keep it as it is now because IMHO it gives actual examples of what you would use Core NATS subject mapping for, and that it is not easy for an application programmer to define his/her own mappings with Core NATS subject mapping as it's an admin operations. Not sure the paragraph really benefits from being shorten to just that one sentence...

adr/ADR-XX.md
jnmoyne: Done

adr/ADR-XX.md
jnmoyne: Done

adr/ADR-XX.md
jnmoyne: Added multiple sources with transform into the example source config JSON

tbeets (approved):
LGTM

PR #205, opened 2023-03-11
add ADR-33 consistant metadata to micro service and endpoints

ripienaar:


ripienaar:
We had some feedback in the new tech previews channel and ADR-33 came later than micro ADR so seems good to have this for consistency.

The use case is that these annotations can be read to build monitoring, ingress and more in a zero config manner.

adr/ADR-32.md
```diff
@@ -123,6 +125,7 @@ Returns a JSON having the following structure:
     name: string,
     id: string,
     version: string,
+    metadata: <string,string>;
```
aricart, line 128: this should be `Record<string,string>`

adr/ADR-32.md
```diff
@@ -194,6 +198,7 @@ Service API does not have any assumptions about API spec or schema formats.
      * The subject on which the endpoint is registered
      */
     subject: string,
+    metadata: <string,string>;
```
aricart, line 201: same as above

adr/ADR-32.md
```diff
@@ -217,6 +222,7 @@ The type for this is `io.nats.micro.v1.schema_response`.
     name: string,
     id: string,
     version: string,
+    metadata: <string,string>;
```
aricart, line 225: same as above

aricart (approved):
LGTM!

bruth (approved):

adr/ADR-32.md
```diff
@@ -172,6 +175,7 @@ Service API does not have any assumptions about API spec or schema formats.
     name: string,
     id: string,
     version: string,
+    metadata: Record<string,string>;
```
scottf, line 178: What is Record<string,string>, sorry not everyone is familiar with this construct.

scottf (changes requested):

adr/ADR-32.md
codegangsta: It's just an associative array in typescript. Aka map, hash, dict etc

ripienaar: Same as stream and consumer metadata

ripienaar: Definitely need to use one language to describe all ADRs :)

ripienaar:
Go PR adding support for this https://github.com/nats-io/nats.go/pull/1231 hope it clarifies

piotrpio (approved):
LGTM

ripienaar:
I think we need to add metadata also to endpoint stats, today its only in schema but in order to build a UI that just shows info about a service we already need to call info (no endpoint details just subjects) and then stats to get endpoint details. As is we would also need to call schema to get endpoint metadata and then a bunch of work to combine this information.

So before I send a updated PR just wanted to solicit feedback on that @piotrpio @aricart

aricart:
@ripienaar the funny thing is - this is what I already did - when you call stats, the `metadata` is included. I think `schema` shouldn't be part of that, as that may be rather verbose.

The part I need to verify on my PR is that `schema` is including `metadata` as well.

ripienaar:
> @ripienaar the funny thing is - this is what I already did - when you call stats, the `metadata` is included. I think `schema` shouldn't be part of that, as that may be rather verbose.
> 
> The part I need to verify on my PR is that `schema` is including `metadata` as well.

You need to be careful about just adding random stuff as the CLI will schema validate responses :)

PR #209, opened 2023-03-17
Adding detail PR 176 js simplification ADR

scottf:
* Changes to defaults are based on conversations with @aricart 
* Captured handling 404's against pending counts
* Suggested warnings to be telegraphed to user.

### DIFF
--- a/adr/ADR-32.md
+++ b/adr/ADR-32.md
@@ -197,10 +197,9 @@ Pending messages and bytes count should be updated when:
 - A new pull request is published - add a value of `request.batch_size` to the pending messages count and
 the value of `request.max_bytes` to the pending byte count.
 - A new user message is processed - subtract 1 from pending messages count and subtract message size from penging byte count.
-- A pull request termination status is received containing `Nats-Pending-Messages` and `Nats-Pending-Bytes` headers, subtract the value of `Nats-Pending-Messages` header from pending messages count and subtract the value of `Nats-Pending-Bytes` from pending bytes count.
+- A pull request termination status is received containing `Nats-Pending-Messages` and `Nats-Pending-Bytes` headers, subtract the value of `Nats-Pending-Messages` header from pending messages count and subtract the value of `Nats-Pending-Bytes` from pending bytes count. Clients could just check all statuses for the headers to future proof.
   - 408 Request Timeout
   - 409 Message Size Exceeds MaxBytes
-  - Others???
 - A pull request termination status of 404 No Messages is received, subtract the entire pull's batch size and max bytes from the pending messages / bytes count.
 
 ###### Message Size Calculation
--- a/adr/ADR-32.md
+++ b/adr/ADR-32.md
@@ -150,10 +150,10 @@ For each constraint, a corresponding threshold can be set.
 > Values used as defaults if options are not provided are subject to further discussion.
 > The values presented below are not definite.
 
-- `max_messages` - 100, [1-???]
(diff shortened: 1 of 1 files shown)


adr/ADR-32.md
```diff
@@ -150,10 +150,10 @@ For each constraint, a corresponding threshold can be set.
 > Values used as defaults if options are not provided are subject to further discussion.
 > The values presented below are not definite.
 
-- `max_messages` - 1000, [1-???]
-- `expires` - 60s, [4s-10m]?
+- `max_messages` - 100, [1-???]
+- `expires` - default 30s, minimum 1s, maximum ???
 - `max_bytes` - not set, use `max_messages` if not provided
-- `idle_heartbeat` - 30s, [2s-60s]?
+- `idle_heartbeat` - 1/2 of expires, minimum 1s, maximum 30s
```
aricart, line 156: @piotrpio had some specific values he was thinking about

adr/ADR-32.md
```diff
@@ -197,18 +197,40 @@ Pending messages and bytes count should be updated when:
...
-
-There are 2 kinds of errors which can be received as a response to a pull request:
-
-- Pull request termination error is any error containing `Nats-Pending-Messages` and `Nats-Pending-Bytes` headers.
-These should not be treated as errors by the application, but rather used to calculate the pending messages/bytes count.
-- Every other pull request error should be treated as terminal error - it should be telegraphed to the user (in langie-specific way),
-the `Consume()` execution should be stopped and subscription should be drained.
+- A pull request termination status is received containing `Nats-Pending-Messages` and `Nats-Pending-Bytes` headers, subtract the value of `Nats-Pending-Messages` header from pending messages count and subtract the value of `Nats-Pending-Bytes` from pending bytes count.
+  - 408 Request Timeout
+  - 409 Message Size Exceeds MaxBytes
```
aricart, line 202: There can be many more 409...

adr/ADR-32.md
```diff
@@ -197,18 +197,40 @@ Pending messages and bytes count should be updated when:
...
+In addition to possibly providing termination `Nats-Pending-Messages` and `Nats-Pending-Bytes` headers, 
+status messages indicate the termination of the pull.
+Statuses that are errors or warnings should be telegraphed to the user in language specific way.
+
+Errors:
+- 400 Bad Request
+- 409 Consumer Deleted
+- 409 Consumer is push based
+
+Warnings:
```
aricart, line 224: None of these need to be telegraphed to the client (not saying whether we should or not, it is just that they are all recoverable.

- if the pull request all the sudden has a message that fits, it will be yielded
- the max request batch is an error, but if you change the consumer configuration, the client will start working
- the request expires is also controlled by configuration on the consumer that can change out of band
- the max waiting will drop off if some messages are acked, etc.

aricart (approved):
Insted of documenting errors that should be ignored, we should focus on the ones that will fail the client, like consumer deleted, or bad argument on the pull, etc. Otherwise we are documenting status messages that the server will send that perhaps should be telegraphed, but if the client has a need for those likely they should be implementing their own protocol.

adr/ADR-32.md
scottf: I just check all status for the headers to future proof the code

adr/ADR-32.md
scottf: I suppose I could change to say Warnings are not required to be telegraphed

PR #216, opened 2023-04-06
Updates ADR with the relaxing of the checking of the republish source against the stream's listened subjects

jnmoyne:


tbeets (approved):
LGTM

PR #217, opened 2023-05-02
ADR-35: JetStream Filestore Compression

neilalexander:
Although mostly done and anticipated to land in 2.10, this ADR describes the filestore compression and the configuration surface area required for clients.


ripienaar (commented):
It might be worth to describe the system behavior when a stream edit enables/disabled compression.

What is the outage users can expect? When will the API call return? How does the user know its completed etc

ripienaar:
@neilalexander do you plan to finish this one? Else lets close it, though I think it has value

PR #227, opened 2023-06-22
ADR-39: Cert Store

tbeets:
Reference documentation for the Cert Store (aka wincert) feature that provides option for NATS Server (on Windows platform) to obtain server's certificate from and delegate handshake signature to the Windows Certificate Store.


wallyqs (approved):
LGTM 👍

PR #230, opened 2023-06-27
ADR-28: RePub, Fix typo in JSON fragment example

tbeets:
The JSON fragment example has `"headers_only": "false"` in example which fails API validation (2.10-Dev).  Should be `"headers_only": false`.   

Added an Updates to ADR-28 section to track.


neilalexander (approved):
LGTM

Jarema (approved):
LGTM

PR #231, opened 2023-07-17
Clarify that when in non-operator mode, the name of the account is not an nkey

aricart:


hspaay:
After some more consideration, I like to recommend a solution described here: https://github.com/nats-io/nats-server/issues/4335  (make config for server and operator mode identical)
This might imply some changes to the document but you'll benefit with reduced support the current configuration will cause.

ripienaar (approved):
LGTM

PR #241, opened 2023-09-19
Update ADR 36

jnmoyne:
Remove references to the single field subject transform destination
Add section about KV bucket sourcing


Jarema (approved):
LGTM!

PR #245, opened 2023-10-11
[EDIT] ADR-32 Add comments around subject

scottf:


adr/ADR-32.md
```diff
@@ -153,7 +153,14 @@ Returns a JSON having the following structure:
 ```typescript
 // EndpointInfo
 {
+    /**
+     * The name of the endpoint
+     */
     name: string,
+    /**
+     * The subject on which the endpoint is registered.
```
aricart, line 161: I would simply say, the subject where the endpoint is listening. - how you got there doesn't matter.

scottf: fixed

aricart (approved):
LGTM!

PR #263, opened 2024-02-01
ADR 41: partitioned consumer groups and exclusive consumer

jnmoyne:
Consistently hashed partitioned consumer groups (including exclusive client consumption from a durable consumer) for streams.


ripienaar:
@jnmoyne can I close this one but keep the branch around for future reference, we'll refine around the new groups ADR and eventually get back to this but I imagine it would be quite a bit rewrite of this ADR then - big enough to really be new?

jnmoyne:
Most of this is still valid and only a small part would have to be re-written with the addition of pinned consumers.

jnmoyne:
Updated to the current initial implementation using the 2.11 features

jnmoyne:
I'll pick the next available ADR number after an LGTM

leonsmith:
Super pumped about this ADR, as this would allow us to move away from sqs with their [MessageGroupId](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/using-messagegroupid-property.html) & finally centralise on NATs

jnmoyne:
Version 0.2.0 now available 
Golang: https://github.com/synadia-io/orbit.go/releases/tag/pcgroups%2Fv0.2.0
Java: https://github.com/synadia-io/orbit.java/releases/tag/pcgcli%2F0.2.0
.Net: https://github.com/synadia-io/orbit.net/releases/tag/PCGroups%2Fv1.0.0-preview.6

PR #269, opened 2024-02-28
Clarify allow_trace behaviors for services and streams

ripienaar:


adr/ADR-41.md
```diff
@@ -77,24 +77,30 @@ Here we set the `msg_trace` configuration for the `A` account, this enables supp
...
   B {
-   exports = [ { allow_trace: true, stream: "nats.add" } ]
+   exports = [ 
+    // on a service the direction of flow is into the exporting
+    // account, so the exporter need to allow tracing
+    { service: "nats.add", allow_trace: true } 
+   ]
+   
+   imports = [
+    // on a stream import the directin of flow is from exporter into
```
bruth, line 96:
```suggestion
    // on a stream import the direction of flow is from exporter into
```

adr/ADR-41.md
```diff
@@ -77,24 +77,30 @@ Here we set the `msg_trace` configuration for the `A` account, this enables supp
...
+    // the importer, so the importer need to allow tracing
+    {stream: {account: user, subject: ticker}, allow_trace: true}
```
bruth, line 98: This should be account `A`, given the above code block? Should the `ticker` export be defined as well?

ripienaar: It could be A, mauby be clearer. 

I am just showing that the thing that needs to be set it only on the importer of a `stream` import, the export doesnt change.

But on a `service` the export change not the import.

So only showing the bits that actually different from typical

bruth: Ok, there is a chance it confuse the average reader, but we can allow a contribution to address if you don't want to fix it now.

bruth (approved):

adr/ADR-41.md
ripienaar: Well fully the idea is that these documents are not for the average reader, we absolultely need to treat these as input for general audience not output.

It's implementers and early adopters and deep technical reference.

PR #279, opened 2024-05-15
ADR-42 consumer priority groups

ripienaar:


ripienaar:
This is a draft, I have some work left to do here and some thinking but enough to get some feedback, i am not feeling too hot today but wanted to get this out, please discuss here.

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+Here we state that the Consumer has one group called `jobs`, it is operating on the `overflow` policy and requires `explicit`
+Acks, any other Ack policy will produce an error. If we force this ack policy in normal use we should error in Pedantic mode.
+
+Pull requests will have the following additional fields:
+
+ * `"group": "jobs"` - the group the pull belongs to, pulls not part of a valid group will result in an error
+ * `"min_pending": 1000` - only deliver messages when `num_pending` for the consumer is >= 1000
+ * `"min_ack_pending: 1000` - only deliver messages when `ack_pending` for the consumer is >= 1000
+
+If `min_pending` and `min_ack_pending` are both given both have to be satisfied (boolean AND).
```
Jarema, line 94: I wonder if this is better, or the way how stream limits works - first threshold passed means overflow condition is satisfied (boolean OR).

I think that in most cases users want threshold condition to kick in, so OR makes more sense, however I can clerly understand that AND could also be useful.

We could reason about it this way:
- OR - we want a looser conditions, that allow overflowing if processing clients are too slow, or they try to process too much at once (max ack pending is high)
- AND - we want be able to tighten the conditions and only overflow if both - clients process too fast/with too much parallelism and are lagging behind.

I think that the second case is rare. Either clients are not catching up, or are trying "too hard" (depending on code and user strategy). Matching both criterias sounds like a "approaching imminent disaster" scenario, where overlow tries to prevent that. Which, when I just said it, actually makes an interesting use case...

ripienaar: I think you're right yeah, will update

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+
+### `pinned_client` policy
+
+Users want to have a single client perform all the work for a consumer, but they also want to have a stand-by client that
+can take over when the primary, aka `pinned` client, fails.
+
+Users would like the ability to influence which client will generally be considered the active, pinned, one - but we feel
```
Jarema, line 101: We agreed why this is the safe idea, and priority can have other uses too, however I wonder if two scenarios will pop up often:

**Users want to stepdown a pinned consumer administratively. Just like stream stepdown.**
With this design, every operation like this would mean lowering the priority. That could lead to priority drifing to 1 which would block the possibility to move further.

Maybe we can have this without any changes in ADR, just some new API call in the future?

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+pinned client:
+
+ 1. Stop delivering messages for this group, wait for all in-flight messages to be completed, continue to serve heartbeats etc 
+ 2. Pick the new pinned client by priority and group
+ 3. Store the new pinned `nuid`
+ 4. Deliver the message to the new pinned client with the ID set
+ 5. Create an advisory that a new pinned client was picked
+ 6. Respond with a 4xx header to any pulls, including waiting ones, that have a different ID set. Client that received this error will clear the ID and pull with no ID 
+
+If no pulls from the pinned client is received within `PriorityTimeout` the server will switch again using the same flow as above.
```
Jarema, line 151: I'm pretty sure rather quickly someone will ask for graceful, scheduled handover of the pinned consumer to not cause any (or almost any) interrupts in opeartions.

Maybe this could be addressed with `consume()` like pattern and very low PriorityTimeout. Or even 0, as `consume()` always has at least one pending Fetch Request.

Jarema (commented):
Some comments and thoughts around details and future proofing.

adr/ADR-42.md
ripienaar: I mention here later a API to trigger a switch - but if people are fiddling priorities too much they'd just end up picking the one that just was the pinned one again. I am kind of less sold on the priority idea atm

adr/ADR-42.md
ripienaar: Yeah, I think we just don't know enough about the patterns users will use this for - if we build a whole solution ontop of this ourselves like the one we are planning we could definitely do that - but I reckon for sute this tunable have to exist I can't imagine a good number for everyone since I suspect people will use these for all sorts of stuff that we can't even anticipate.

adr/ADR-42.md
ripienaar: I might not add the priority idea in the initial build I think, what do you think @Jarema gives us time to think about this more. I'm definitely also worried about lots of unexpected switches of pinned client if someone just gets the timings a bit wrong and priorities will make that worse

Jarema: Yes, I agree. 
I really like that we're designing the features with future in mind, but we also do not have to implement those part of the design that are not required for first steps.

Addition of priority later should also not break anything, as we could say that not setting priority sets its to, for example, 5.

ripienaar: @jnmoyne would you say being able to influence next-pinned-client by priority is important for initial work? We want some more time to explore how that will work

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+to ensure serving client needs as soon as possible.
+
+The `overflow` policy enables Pull requests to be served only if criteria like `num_pending` and `num_ack_pending` are
+above a certain limit, otherwise those Pull requests will sit idle in the same way that they would if no messages were
+available (receiving heartbeats etc).
+
+```go
+{
+	PriorityGroups: ["jobs"],
+	PriorityPolicy: "overflow",
```
aricart, line 79: This should also be an option on regular consumers

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+ * `"group": "jobs"` - the group the pull belongs to, pulls not part of a valid group will result in an error
+ * `"min_pending": 1000` - only deliver messages when `num_pending` for the consumer is >= 1000
+ * `"min_ack_pending: 1000` - only deliver messages when `ack_pending` for the consumer is >= 1000
+
+If `min_pending` and `min_ack_pending` are both given either being satisfied will result in delivery (boolean OR).
+
+### `pinned_client` policy
+
+Users want to have a single client perform all the work for a consumer, but they also want to have a stand-by client that
+can take over when the primary, aka `pinned` client, fails.
```
aricart, line 99: "high-availability" consumer?

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+ 1. Stop delivering messages for this group, wait for all in-flight messages to be completed, continue to serve heartbeats etc 
+ 2. Pick the new pinned client by priority and group
+ 3. Store the new pinned `nuid`
+ 4. Deliver the message to the new pinned client with the ID set
+ 5. Create an advisory that a new pinned client was picked
+ 6. Respond with a 4xx header to any pulls, including waiting ones, that have a different ID set. Client that received this error will clear the ID and pull with no ID 
+
+If no pulls from the pinned client is received within `PriorityTimeout` the server will switch again using the same flow as above.
+
+Clients can expose call-back notifications when they become pinned (first message with `Nats-Pin` header is received) and
```
aricart, line 153: `Nats-HA`

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+when they lose the pin (they receive the 4xx error when doing a pull with a old ID).
+
+A new API, `$JS.API.CONSUMER.UNPIN.<STREAM>.<CONSUMER>.<GROUP>`, can be called which will clear the ID and trigger a client switch as above.
```
aricart, line 156: `SELECT` - pin is weird to me

adr/ADR-42.md
```diff
@@ -0,0 +1,168 @@
...
+
+Consumer state to include a new field `PriorityGroups` of type `[]PriorityGroupState`:
+
+```go
+type PriorityGroupState struct {
+	Group          string `json:"name"`
+	PinnedClientId string `json:"pinned_id,omitempty"`
```
aricart, line 163: `SelectedClientId`
`LeaseTs`

aricart (approved):
LGTM - I put some suggestions on terms/option names

adr/ADR-42.md
ripienaar: I thought about that, but I really dont think we know all the implications - like for example just on the other policy we'd have to then error if these are added.

Seems like a can of worms of crazy matrics of features.  So I am really keen to move to a more clear intent when it comes to these things.  You want this behavior you gotta say so rather than depend on a 5 boolean if statement :)

adr/ADR-42.md
ripienaar: its definitely not selecting, its more clearing the pin and then later doing a switch.

We had a lot of to and fro on the name and `pin` really worked for most people :)

adr/ADR-42.md
```diff
@@ -0,0 +1,161 @@
...
+### `overflow` policy
+
+Users want certain clients to only get messages when certain criteria are met.
+
+Imagine jobs are best processed locally in `us-east-1` but at times there might be so many jobs in that region that 
+`us-west-1` region handling some overflow, while less optimal in terms of transit costs and latency, would be desirable 
+to ensure serving client needs as soon as possible.
+
+The `overflow` policy enables Pull requests to be served only if criteria like `num_pending` and `num_ack_pending` are
+above a certain limit, otherwise those Pull requests will sit idle in the same way that they would if no messages were
```
neilalexander, line 73: At the moment I'm trying to think through how this would be implemented. The `num_ack_pending` would only be observable on the same node that the first consumer is running on, and I'm also wondering if there might also be issues with fetching `num_pending` from non-stream leaders if the replicas are behind but the second consumer _should_ be kicking in by that point creating a gap.

jnmoyne (approved):
Just some minor test edit suggestions, and some comments (things to be captured for when the documentation is going to be written based on this document). LGTM.

adr/ADR-42.md
jnmoyne: No, for initial work I would want just pick any other available. People that are asking for priority would be satisfied with 'overflow'. IMHO trying to have both priority an exclusivity is an architectural mistake. If you do it, the next thing they will ask for is that if the pinned client has moved to a lower priority one because the higher priority one has left, then when one with a higher priority comes back up they will want it become again the new pinned client.
Otherwise you end up with the priority race condition you describe above. And if you do that then it is effectively 'overflow' all over again.

So to do 'exclusive with priority' you just use 'overflow' with just one instance per priority level (should be trivial to do with any container orchestration), therefore I don't think we would ever need to add priorities to pinned.

adr/ADR-42.md
```diff
@@ -0,0 +1,165 @@
...
+	AckPolicy: "explicit",
+	// ... other consumer options
+}
+```
+
+This configuration states:
+
+ * We have 1 group defined and all pulls have to belong to this group
+ * The policy is `pinned_client` that activates these behaviors
+ * When a pinned client has not done any pulls in the last 120 seconds the server will switch to another client
```
Jarema, line 124: With this approach, longer expire interval, more sensitive the TTL becomes.
Maybe we should base TTL on Duration since last Pinned Client Pull Request expiry?

adr/ADR-42.md
```diff
@@ -0,0 +1,165 @@
...
+In the specific case where MaxAckPending is 1 and a pull is made using `min_pending: 1` this should only be served when
+there are no other pulls waiting. This means we have to give priority to pulls without conditions over those with when
+considering the next pull that will receive a message.
+
+### `pinned_client` policy
+
+Users want to have a single client perform all the work for a consumer, but they also want to have a stand-by client that
+can take over when the primary, aka `pinned` client, fails.
+
+**NOTE: We should not describe this in terms of exclusivity as there is no such guarantee, there will be times when one
```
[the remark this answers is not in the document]
jnmoyne: I use double acks in my example for consumer groups for that very reason (i.e. if the double ack returns an error then 'rollback' rather than 'commit')

adr/ADR-42.md
jnmoyne: FWIW in my initial consumer group implementation I am using the idle timeout for the consumer (i.e. time since last pull request expired) and it's working for me. Otherwise what if you set the timeout to for example 60s and the applications are fetching with a timeout of more than 60s, wouldn't that lead to flapping when there are no new messages to deliver?

ripienaar: Maybe not clear here but like the way it calculates if a consumer is idle or not it wouldn’t consider a client with an flight pull as being active. 

But this seems open to bugs or abuse so perhaps you simply would have to ensure you pull more frequently. Alternatively we could cap the pull timeout so it relates to this timeout - no more than twice this timeout pull expire or something?

good point. Needs some thinking here

adr/ADR-42.md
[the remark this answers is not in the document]
Jarema: @jnmoyne how do you think double ack help you here?
Double acking will not even return error if you ack outside of the ack_wait window, or done from different puller.

AFAIK server never sends back an error in double ack. It's the client timing out/getting no responders being the only cases for errors.

There is a related issue here: https://github.com/nats-io/nats-server/issues/5113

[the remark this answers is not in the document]
jnmoyne: This is the failure scenario I could reproduce with my consumer group code and example: the active fetcher gets a message (in my example max acks pending is 1, as it's kind of the whole point with consumer groups :)) and gets suspended while processing the message (similar to if it gets isolated because for example of a network split-brain), while it is suspended another client becomes the active fetcher and start receiving (and processing) messages, starting with the one message that had been delivered once to the previous (now suspended) active client. So the new guy gets that message and processes it and (double) acks it and it all goes well, and it goes on to receive and process further messages. Then the previous active client gets un-suspended continues processing its message and then acks it (though it was actually processed and acked by the new active client in the mean time), which is a problem. 

So what I show in my example is how (if your processing at the end involves writing the result of the processing to a transaction-capable system like an SQL DB for example) you can avoid that problem and get 'exactly once processing'. What the client code does is: upon receiving the message is starts the transaction, then does it processing and prepares the transaction. At that point it does a double-ack on the message, if the message had been acked before by someone else (i.e. the client that became active while this one was suspended or isolated) then the double-ack fails and it does a rollback, if the double ack returns successfully then it does a commit instead.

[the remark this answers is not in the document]
Jarema: Server **will not** send you back error if you ack the same message second time.

Client might check for double acking the same message twice, but that's it. It would not work across clients/connections.

Hence you cannot rely on that mechanism.

In fact, server **never** responds with an error, unless it's `no responders`, which happens only if jetstream is disabled.

You can even mess up stream sequences, delivert sequences etc, and server will happily respond with ok.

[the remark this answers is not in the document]
jnmoyne: In my experimentation the double ack would return an error when I caused another client to become active (and therefore get re-delivery of the message and acking it) and it gets un-suspended and tries to double ack the same message.

Never realized that the server did not return an error if the message is already acked, and it certainly fooled me into thinking that the server did that.

In any case I would consider it a bug that the server doesn't return an error to an ack request if the message in question has already been acked. To me that's a bit the whole point of a double ack (besides knowing that the ack message has made it to the consumer's leader): knowing that the ack was indeed successful or not, and I would consider trying to ack an already ack'd message as an error (I would even go as far as expecting an error when trying to ack with a delivery count that doesn't match the current delivery count for the message).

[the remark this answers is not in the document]
ripienaar: Would be interested in knowing what error you get, anyone can ack any messages - delivered to that client or not, already acked or not, ack wait expired or not etc.

It's an area of big concern for exactly once behaviors imo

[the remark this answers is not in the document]
Jarema: @ripienaar Go client tracks if given message was already acked, and will error if you ack it second time. Obviously works only for acks done by the same client instance. To be fair - I dont like that behavior considering how little info you actually get from the server and that it might be necessary to ack it again in some failure scenarios that server did not inform about.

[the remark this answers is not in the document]
ripienaar: Oh yes that's definitely not behaviour we need in the client imo, it will be confusing as we saw but also leed to a false sense of whats possible when things only work within one client connection.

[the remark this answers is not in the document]
jnmoyne: I checked and the actual error I get is a 'no responders available' - probably more an effect of the way I trigger the failure scenario. 

In any case I agree (I think we all agree) that this current behavior of not returning an error if already acked (or allowing to ack a message not even delivered) is not right and does have implications for 'exactly once' behaviors.

adr/ADR-42.md
```diff
@@ -0,0 +1,166 @@
...
+   this field is absent
+
+After selecting a new pinned client, the first message that will be delivered to this client, and all future ones, will 
+include a Nats-Pin-Id: xyz header. The client that gets this message should at that point ensure that all future pull 
+requests have the same ID set.
+
+When a new pinned client needs to be picked - after timeout, admin action, first delivery etc, this process is followed:
+
+ 1. Stop delivering messages for this group, wait for all in-flight messages to be completed, continue to serve heartbeats
+ 2. Pick the new pinned client by rtt
```
Jarema, line 140: After call with Derek, we learned that the RTT we have access to in the server is not reliable and we should just pick the first available Request.

Jarema: AFAIK Stream Alternates are sorted by Gateways RTT, not client RTT.

adr/ADR-42.md
```diff
@@ -0,0 +1,166 @@
...
+ * When a pinned client has not done any pulls in the last 120 seconds the server will switch to another client
+ * AckPolicy has to be `explicit`. If we force this ack policy in normal use we should error in Pedantic mode
+
+A pull request will have the following additional fields:
+
+ * `"group": "jobs"` - the group the pull belongs to, pulls not part of a valid group will result in an error
+ * `"id": "xyz"` - the pinned client will have this ID set to the one the server last supplied (see below), otherwise 
+   this field is absent
+
+After selecting a new pinned client, the first message that will be delivered to this client, and all future ones, will
```
Jarema, line 133: There are two things to discuss here:
1. This will be the first time we send status/api-like header on an actual user message. Nothing bed in it. Just mentioning that this is a new pattern.
All other headers set in msgs by the server are informational, not status-like.
2. Sending the header in every message will have some overhead. I wonder if its not enough to sent the initial message. After all, sending the wrong along the pull request will trigger a server error message.

adr/ADR-42.md
```diff
@@ -0,0 +1,166 @@
...
+We introduce 2 settings on `ConsumerConfig` that activates these related features:
+
+```go
+{
+	PriorityGroups: ["<group>", ...],
+	PriorityPolicy: "<policy>"
+}
+```
+
+The presence of the `PriorityGroups` setting activates the set of features documented here with a list of names supported
```
Jarema, line 49: After working with the codebase, I came to conclusion that making `PriorityGroups` an activator makes it really unintuitive from client perspective and not very beneficial for future proofing. 
Why client needs to set `PriorityGroups`, which does nothing right now, if they can just set `PriorityPolicy` which is what makes sense to actually check in the code?

I absolutely support the idea of thinking forward with the how we will expand the feature set, I'm just not convinced this is the way to go for above reason and few others:
1. Are we confident that each `PriorityGroup` cannot have it's own `PriorityPolicy`? If yes, `PriorityGroups` should be rather `PriorityGroups: []PriorityGroup`, which would contain group name, policy, etc. Otherwise, we would probably anyway introduce changes that are hard to not make them breaking, or easy to work with.
2. Handlin many priority groups requires a pretty heavy refactor of how we fetch messages. That can impact the API too in a ways we have not yet discovered.

I'm a bit afraid of having confusing API saying "if you want to use `PriorityPolicy`, you need to also specify `PriorityGroups`. It's a slice, but you can add only one member. And it does not matter now."

On the other hand, picking default name for priority groups could work (if `PriorityPolicy is set, and `PriorityGroups` are not, set a a default `PriorityGroups`), but that again will be a usecase for pedantic mode, and can become surprising when we will start actually support `PriorityGroups`.

**I think we should either make `PriorityGroups: []PriorityGroup` straigtht away, and say that only one can be passed as a compromise between future proofing and good ux, or remove the requirement.**

ripienaar: Group name is used in pinned clients.  Using it now on day one even though a bit odd being just one means old clients will remain forward compatible once we support more or add meaning to those group names.  They will already have a group name set and continue to work.

On the other hand if we drop the group name requirement and only introduce it later we get a cliff where all older clients just start erroring.  Perhaps thats a good forcing function but I suspect what we'd end up doing in that case is further complicate the server code to try and handle the case of no group and multiple groups - at that point not having set it today becomes the worst possible outcome.

Jarema: Sure, that is right, IF we are 100% sure that we will not want PriorityGroups with different settings each - so `PriorityGroups: []string` vs `PriorityGroups: []PriorityGroup`.

I dont have strong opinion here - I agree with you. But I think that if we want to introduce groups as mandatory public API, we need to think and be confident which of above cases will be future proof.

ripienaar: We can certainly explore if the list of policy is the right approach - I agree a map might be better long run.  

I just know what we'll do when the cliff comes - we'll complicate the server to deal with our earlier minimal API design rather than force upgrades so I think the effort to explore some possible futures a bit more would be worth it.

ripienaar: We discussed this and thought making groups maps would get us into a complexity trap very quickly.

We'll keep groups as a requirement - but limit it to just one for now.

You say presence of groups enables the feature or setting a priority policy its functionally equivelant since a group is required.  I'll update the doc to say the policy activates it - but since both are required its functionally equiv

adr/ADR-42.md
ripienaar: Regarding when we send the header, I did that to guard against the first message being a magical message that can never be lost which would lead to a reselection happening after timeout. While if its on each message the first can be lost and things will still work - that first will just later be retried.

Seems worth it to me, wdyt @Jarema

Jarema: Yes, I agree with that. However I still do not like the overhead. What about sending the `Nats-Pin-Id` in each message in first batch? 
Missing whole batch might suggest anyway a problem with a client. But yeah, there is also a batch of one...

Ok, let's stich to sending it in every message.
We can always revisit.

ripienaar:
@Jarema see https://github.com/nats-io/nats-architecture-and-design/pull/279/commits/7250d66cc01131b570ad204735b047ddde8149a3 to summarise our call about config updates

Jarema (approved):
LGTM

PR #292, opened 2024-07-11
ADR-43: message ttl

ripienaar:


adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+We could optionally extend the `PubAck` as follows:
+
+```golang
+type PubAck struct {
+	MsgTTL    uint64 `json:"msg_ttl,omitempty"`
+}
+```
+
+This gives clients a chance to confirm, without Stream Info or should the Stream be edited after Info, if the TTL
```
Jarema, line 66: I agree with the reasoning and removing uncertaintity, but this means that because server cannot tell us if it does support given feature, we're having two mechanisms to confirm ack ttl issue - API error (for server supporting `Nats-TTL`) and one for server not supporting it - returning struct without `MsgTTL` field. The concern is that **every** ack will be  bigger by a few bytes, even years in the future, where all streams will support this feature.

Jarema (commented):
Looks good! Some comments and thoughts added.

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+We will allow a message to supply a TTL as a duration in a header called `Nats-TTL` followed by the duration as 
+nanoseconds (TODO: or seconds? I prefer seconds but consistency suggests ns).
+
+The duration will be used by the server to calculate the deadline for removing the message based on its Stream 
+timestamp and the stated duration.
+
+The TTL may not exceed the Stream MaxAge to give administrators a hard cap for the purposes of managing resources.
+
+A TTL of zero or other unparsable value will result in a error (TODO: or ignore? I prefer error) and reported in the
```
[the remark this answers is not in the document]
ripienaar: Yeah sounds good, will update

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+The duration will be used by the server to calculate the deadline for removing the message based on its Stream 
+timestamp and the stated duration.
+
+The TTL may not exceed the Stream MaxAge to give administrators a hard cap for the purposes of managing resources.
+
+A TTL of zero or other unparsable value will result in a error (TODO: or ignore? I prefer error) and reported in the 
+Pub Ack.
+
+Any value less than 0 will result in a message being retained forever, even if the stream have a MaxAge set.
```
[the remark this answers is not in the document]
ripienaar: Sure but we wouldnt expose it as a duration perhaps? Like

```
kv.Put("x", "y", TTL(1h))
```

vs

```
kv.Put("x", "y", NoExpire())
```

under the covers your type is a int64?

Anyway, I am not wed to the specific approach, happy with an extra header also.

adr/ADR-43.md
ripienaar: That then leaves the 2 questions:

 * How do you know if you can use this feature - do we require a stream info always? Even for KV?
 * What if someone updates the stream since your info - kv handles can live a very long time in long running processes

adr/ADR-43.md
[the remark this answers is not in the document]
ripienaar: Added `Nats-No-Expire`

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+Streams support a one-size-fits-all approach to message TTL based on the MaxAge setting. This causes any message in the 
+Stream to expire at that age.
+
+There are numerous uses for a per-message version of this limit, some listed below:
+
+ * KV tombstones are a problem in that they forever clog up the buckets with noise, these could have a TTL to make them expire once not useful anymore
+ * Server-applied limits can result in tombstones with a short per message TTL so that consumers can be notified of limits being processed. Useful in KV watch scenarios being notified about TTL removals
+ * A stream may have a general MaxAge but some messages may have infinite retention, think a schema or type hints in a KV bucket that is forever while general keys have TTLs
+ * TODO: user use cases
```
ripienaar, line 20: Please help me with links to some user use cases from issues and elsewhere if you have any @bruth ?

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+This behaviour is off by default unless opted in on the Stream Configuration.
+
+## Publish Acknowledgements
+
+We could optionally extend the `PubAck` as follows:
+
+```golang
+type PubAck struct {
+	MsgTTL    uint64 `json:"msg_ttl,omitempty"`
```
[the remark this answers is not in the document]
ripienaar: There isnt a guarantee the server will expire the message at any exact deadline time stamp so I was reluctant to put in a timestamp here.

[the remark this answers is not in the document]
ripienaar: Also no guarantee about the server time vs client time, so unless the puback also included the server timestamp it still would be an unknown.  I've wanted to include the ts in the ack but suspect it will be too many bytes for not a lot of utilty.

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+ * A stream may have a general MaxAge but some messages may have infinite retention, think a schema or type hints in a KV bucket that is forever while general keys have TTLs
+ * TODO: user use cases
+
+Related issues [#3268](https://github.com/nats-io/nats-server/issues/3268)
+
+## Per-Message TTL
+
+We will allow a message to supply a TTL as a duration in a header called `Nats-TTL` followed by the duration as 
+nanoseconds, the sever may not support nanosecond level precision and clients can choose to expose TTL as
+seconds or nanoseconds.
```
aricart, line 28: nano seconds is overkill - should be seconds or millis at most.

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+The duration will be used by the server to calculate the deadline for removing the message based on its Stream 
+timestamp and the stated duration.
+
+The TTL may not exceed the Stream MaxAge to give administrators a hard cap for the purposes of managing resources.
```
aricart, line 33: Actually - you could from the description earlier - it would be useful to have a TTL that is beyond the max age, that way you can have a permalink message, whether others are pruned as the lifecycle continues. Sounds non-intuitive, but allows default (no headers set) to clean the messages as they age, but the explicit ones remain.

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+A TTL of zero will be ignored, any other unparsable value will result in a error reported in the Pub Ack and the message
+being discarded.
+
+Setting the header `Nats-No-Expire` to `1` will result in a message that will never be expired.
```
aricart, line 38: was going to say that it would be useful to add messages that don't expire or expire beyond the default expire of "normal" messages expired by the TTL in the stream - perhaps it makes sense to allow expirations that are further out from the server TTL or not expire at all.

ripienaar: I was also not sure about this so opted for the safest option for now - that MaxAge is a cap except for the case of no expiry of some messages.

Can be convinced eitherway hence my asking for some real world use case feature requests.  I do think we need a safety feature - which you can disable by just not setting a MaxAge

adr/ADR-43.md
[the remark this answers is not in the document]
aricart: possibly it should be a concrete timestamp in the future. If the data is invalid after a certain amount of time, it is certainly invalid if the server was not able to remove it because it was down, but came up.

adr/ADR-43.md
ripienaar: I think we all agree, but all durations are nanos so I think it best to be consistent - even if we never can get that resolution in precision for this feature.

adr/ADR-43.md
```diff
@@ -0,0 +1,87 @@
...
+
+```golang
+type StreamConfig struct {
+	// AllowMsgTTL allows header initiated per-message TTLs
+	AllowMsgTTL bool          `json:"allow_msg_ttl"`
+
+	// LimitsTTL activates writing of messages when limits are applied with a specific TTL
+	LimitsTTL   time.Duration `json:"limits_ttl"`
+}
+```
```
aricart, line 87: The above is the reason why the capability is on the server, and not a config on the stream. During the lifecycle of the client the stream may change configuration, so now the stream has a config that is different than what was read...

adr/ADR-43.md
ripienaar, line 33: @derekcollison I do not know how you'll implement this but I do want to cap both the minimum TTL and Maximum TTL here to avoid issues of 10 year TTLs might cause.

So once you are a bith further down the implementation detail please let me know what to put here.  I reckon smallest minimum might be seconds and the largest days but open to feedback here based on what you are thinking implementation wise.

derekcollison: Impl will be on second resolution. You can give us anything but that is granularity we will fire.

For maximum, not sure we need one.

adr/ADR-43.md
ripienaar: This is on the list for issues to work on soon, there's now enough buy in and acknowledgement that this should be addressed at a level that spans all the times we've run into this - so I agree but not for this ADR

adr/ADR-43.md
ripienaar: I'll stick with being consisent then and go for nanos - clients can present it however they like.

You're 100% sure we dont need to be cautious here that if someone makes a 100 million message stream all with 10 year TTL that this would not cause issues? For sure I'll cap it at MaxAge or forever but just want to be 100% sure we're not creating another point of abuse.

adr/ADR-43.md
ripienaar: I believe we'll need to support enabling and disabling this at different points in a replication chain so that some downstream source dont start expiring messages if thats not required.

I think we need to consider even removing these headers once ingested - along with a few others thats caused issues with replication - but so far there's not been much appetite for that idea and always reasons to keep them downstream.  So the most feasible is a boolean as here and this config item since its not once size fits all - though I did consider defaulting to MaxAge maybe

adr/ADR-43.md
derekcollison: The abuse will not be in TTL IMO, also we will allow `does not expire` so. The abuse will come in having millions of these in a stream IMO.

ripienaar: OK, made the header second and kept the config option as a duration for consistency.  Minimum for TTL is 1 and uncapped max.

adr/ADR-43.md
```diff
@@ -0,0 +1,85 @@
...
+The duration will be used by the server to calculate the deadline for removing the message based on its Stream 
+timestamp and the stated duration.
+
+The TTL may not exceed the Stream MaxAge to give administrators a hard cap for the purposes of managing resources. 
+The shortest allowed TTL would be 1 second.
+
+A TTL of zero will be ignored, any other unparsable value will result in a error reported in the Pub Ack and the message
+being discarded.
+
+Setting the header `Nats-No-Expire` to `1` will result in a message that will never be expired.
```
Jarema, line 36: It might be a good idea to have a setting enabling/disabling `Nats-No-Expire` to give administrators control over abusing it.

adr/ADR-43.md
```diff
@@ -0,0 +1,85 @@
...
+We have to assume someone will want to create a replication topology where at some point in the topology these tombstone
+type messages are retained for an audit trail. So a Stream with this feature enabled can replicate to one with it 
+disabled and all the messages that would have been TTLed will be retained.
+
+```golang
+type StreamConfig struct {
+	// AllowMsgTTL allows header initiated per-message TTLs
+	AllowMsgTTL bool          `json:"allow_msg_ttl"`
+
+	// LimitsTTL activates writing of messages when limits are applied with a specific TTL
```
Jarema, line 82: So this is another configurable threshold below MaxAge?

Also - consider having setting for disabling `Nats-No-Expire`.

Jarema (changes requested):
Just few last comments

adr/ADR-43.md
ripienaar: When the server places a tombstone after removing the last message for a subject you would need a TTL ok that tombstone but what is the right setting? I think MaxAge but had some disagreement around that so this setting controls the TTL of MaxAge tombstones.

ripienaar: Disabling is a good idea but let’s leave it for later after feedback to get the core behaviour out.

Jarema: @ripienaar Yes, it makes sense for tombstone to not live as long as MaxAge. I understant you name it in a way generic enough that we can reuse it for other purposes, however in this case, I wonder if its not better to just call it `TombstoneTTL`. Though I'm not convinced myself.

ripienaar: We have client added tombstones and server ones and maybe eventually server ones for multiple reasons - so these are the ones added by the stream limits.

Jarema (approved):
LGTM!

aricart (approved):
LGTM

adr/ADR-43.md
bruth: This seems to contradict the above statement:

> The TTL may not exceed the Stream MaxAge to give administrators a hard cap for the purposes of managing resources.

If no `MaxAge` is set on the stream, then _not_ setting the `Nats-No-Expire` would mean is never expires. If a `MaxAge` is set then `Nats-No-Expire` should not be allowed since that violate the hard cap.

ripienaar: We specifically want the ability to keep messages forever.  Imagine a KV where you need to recover the data type for messages in the stream but want the data in general to expire.

These configuration style messages must persist forever while the bulk of the data is limits managed.

As @Jarema says we might need to make this uncapped special messages a boolean but I think we can do that later once we get some feedback

ripienaar:
Have changed the wording around MaxAge rather as a admin control for resource management its stated as the default when not set on a message and also the maximum allowed except for the never expire ones.

We can later restrict that with a config bool if we want but I dont want to add 3 new stream configs right at the start if it can be avoided or later end up not being needed

I'll merge this one end of day baring other feedback.

PR #293, opened 2024-07-17
Adds a deprecated status

ripienaar:
Update some dependencies.


ripienaar:
When deprecating a ADR in favor of some other we can list the final entry in the changelog for the ADR and refer to the new ADR there.

Jarema (approved):
LGTM!

PR #296, opened 2024-07-24
TLS handshake supported server version

roeschter:
TLS handshake supported server version - Looked like it would only be supported for teh upcoming 2.11. But going back 2.10 releasese its obvious the meant to say 2.10.11


Jarema (changes requested):
Good catch! Though it's actually available since 2.10.4
https://github.com/nats-io/nats-server/releases/tag/v2.10.4

Jarema (approved):
LGTM!

PR #300, opened 2024-08-08
ADR-32: Adds a way to make queuegroup-less endpoints

ripienaar:
Also formally ensure service metadata is immutable.


Jarema (approved):
LGTM

PR #318, opened 2025-01-22
Normalize links to other ADR documents

bannatech:
Links to other ADR documents were normalized based off of relative directories which GitHub supports:

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes#relative-links-and-image-paths-in-readme-files

### DIFF
--- a/adr/ADR-10.md
+++ b/adr/ADR-10.md
@@ -22,7 +22,7 @@ the following JSON:
 }
 ```
 
-The `error` field is an [ApiError](0007-error-codes.md). The `success` field will be set to `true` if the request
+The `error` field is an [ApiError](ADR-7.md). The `success` field will be set to `true` if the request
 succeeded. The `purged` field will be set to the number of messages that were
 purged from the stream.
 
--- a/adr/ADR-2.md
+++ b/adr/ADR-2.md
@@ -134,7 +134,7 @@ Advisories must include additional fields:
 Any `message` can have an optional `error` property if needed and can be specified in the JSON Schema,
 they are not a key part of the type hint system which this ADR focus on.
 
-In JetStream [ADR 0001](0001-jetstream-json-api-design.md) we define an error message as this:
+In JetStream [ADR 0001](ADR-1.md) we define an error message as this:
 
 ```
 {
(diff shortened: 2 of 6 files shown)


bruth (approved):

.readme.templ
```diff
@@ -25,6 +25,6 @@ We want to move away from using these to document individual minor decisions, mo
 
 ## Template
 
-Please see the [template](adr-template.md). The template body is a guideline. Feel free to add sections as you feel appropriate. Look at the other ADRs for examples. However, the initial Table of metadata and header format is required to match.
+Please see the [template](./adr-template.md). The template body is a guideline. Feel free to add sections as you feel appropriate. Look at the other ADRs for examples. However, the initial Table of metadata and header format is required to match.
```
ripienaar, line 28: curious why specifically `./x`, when relative worked fine before?

bannatech: `./` is that it is a tad more expressive, and GitHub will still link properly[^1]. 

 [^1]: GitHub even scrubs `./` from the URL...

 `./` is more easily searched for when relative links containing files in the same directory are ever moved into other locations. e.g. if a particular ADR, or all ADRs, are moved to their own sub-directory.  

I am entirely happy with removing `./` from all links if this reasoning doesn't merit the change, but I think there are some upsides to using `./`, from a maintainability perspective.

ripienaar: These are fine examples of why, but no contributor will do this as its unusual :) I prefer to stick to normal paths that is likely to match what contributors would also do

bannatech: That is fair, I'll be glad to rewrite the commit to omit `./`. I don't think any contributor even had a common method for linking other ADRs, as it was a bit all over the place - so people might end up linking to a GitHub page in the end after all :rofl:

ripienaar: Indeed many ways, but none of them ever used `./` which suggests it's pretty certain they won't in future either :)

ripienaar:
If you're feeling generous here are a few other broken things.

 - ADR-10.md links to 0007-error-codes.md which I suppose should be ADR-7.md
 - ADR-2.md links to 0001-jetstream-json-api-design.md which should be ADR-1.md

bannatech:
> If you're feeling generous here are a few other broken things.
> 
> * `ADR-10.md` links to `0007-error-codes.md` which I suppose should be `ADR-7.md`
> * `ADR-2.md` links to `0001-jetstream-json-api-design.md` which should be `ADR-1.md`

Ah, these flew right past me - I've fixed those links too. Thanks!

ripienaar:
Thank you @bannatech

PR #336, opened 2025-04-09
KV max_age vs duplicate_window

scottf:


adr/ADR-8.md
```diff
@@ -21,7 +21,8 @@
 | 6        | 2024-06-05 | Add Keys listers with filters                       |                    |
 | 7        | 2025-01-23 | Add Max Age limit Markers, remove non direct gets   | 2.11.0             |
 | 8        | 2025-02-17 | Add Metadata                                        | 2.10.0             |
-
+| 9        | 2025-04-09 | max_age and duplicate_window notes                  | 2.10.0             |
+| 8a      | 2025-02-17 | Add Metadata | | 2.10.0 |
```
ripienaar, line 25: There's a extra line here

adr/ADR-8.md
```diff
@@ -309,7 +310,11 @@ A bucket is a Stream with these properties:
    * The minimum allowed size is 1. When creating a stream, 1 should be used when the user does not supply a value.
  * Safe key purges that deletes history requires rollup to be enabled for the stream using `rollup_hdrs`
  * Write replicas are File backed and can have a varying R value
- * Overall Key TTL is managed using the `max_age` key
+* Overall Key TTL is managed using the `max_age` key
```
ripienaar, line 313: Formatting is out of line with the rest here - think you need a space infant

ripienaar (changes requested):
Few minor edits, but content seems great thanks

adr/ADR-8.md
scottf: fixed

adr/ADR-8.md
scottf: ugh. fixed

adr/ADR-8.md
```diff
@@ -310,6 +310,10 @@ A bucket is a Stream with these properties:
  * Safe key purges that deletes history requires rollup to be enabled for the stream using `rollup_hdrs`
  * Write replicas are File backed and can have a varying R value
  * Overall Key TTL is managed using the `max_age` key
+    * If Key TTL is supplied (greater than zero), the client should set `duplicate_window` like so:
+        1. if `max_age` is greater than 2 minutes, `duplicate_window` must be set to 2 minutes.
+        2. if `max_age` is less than or equal to 2 minutes, `duplicate_window` must be set the same as `max_age`
+    * If Key TTL is not supplied, is acceptable to either not set `duplicate_window` or set it to 2 minutes. The server will set it to 2 minutes if not supplied.
```
mtmk, line 316: is this regardless of what `max_age` value is?

ripienaar: dupe window is always set yes, only relates to max age when max age is less than dupe window

ripienaar: I am pretty sure the server could remove this restriction all together - but cant exactly change this stuff soon anyway so best we do client fix

PR #341, opened 2025-04-29
[EDIT] ADR-22 Integration and Orbit notes

scottf:


wallyqs:
in the sense that if you want to prototype the approach in orbit, then once got it right move it to the nats.java, yes it would be better to start off in orbit, but eventually nats.java should have it

adr/ADR-22.md
```diff
@@ -16,6 +21,10 @@ try to mitigate these failures, retries can be added into JetStream
 enabled clients to attempt to publish the message to JetStream once it
 is ready again.
 
+#### Client Integration
+
+Although an implementation is provided directly in the GO client, it is preferred that this functionality is built in Orbit or similar, as an extension/library.
```
alexbozhenko, line 26:
```suggestion
Although an implementation is provided directly in the GO client, it is preferred that this functionality is built in [Orbit](https://www.synadia.com/blog/orbit-and-the-future-of-the-clients) or similar, as an extension/library.
```

PR #348, opened 2025-06-13
ADR-50 refinements

ripienaar:


ripienaar:
We leave this one open for a while so further refinements can be added here

PR #352, opened 2025-06-19
Adds the ability to direct get without headers

ripienaar:


adr/ADR-52.md
```diff
@@ -0,0 +1,32 @@
...
+|----------|-------------------------------------|
+| Date     | 2025-06-19                          |
+| Author   | @ripienaar                          |
+| Status   | Implemented                         |
+| Tags     | jetstream, client, refinement, 2.12 |
+| Updates  | ADR-31                              |
+
+# Context
+
+Often the only part of a message users care for is the, a good example is counters introduced in ADR-49 where there is a data section in the body and a control section in the headers. For clients that only care for the current count there is no need to even download the headers from the server.
```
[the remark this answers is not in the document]
ripienaar: yup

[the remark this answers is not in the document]
ripienaar: yup

adr/ADR-31.md
```diff
@@ -84,11 +85,13 @@ StartTime    *time.Time `json:"start_time,omitempty"`
 MultiLastFor []string   `json:"multi_last,omitempty"`
 UpToSeq      uint64     `json:"up_to_seq,omitempty"`
 UpToTime     *time.Time `json:"up_to_time,omitempty"`
+NoHeaders    bool       `json:"no_hdr,omitempty"`
 ```
 
 Example request payloads:
 
 - `{seq: number}` - get a message by sequence
+- `{seq: number, no_hdr: true}` - get a message by sequence, no user headers are included, only the payload
```
MauriceVanVeen, line 94:
```suggestion
- `{seq: number, no_hdr: true}` - get a message by sequence, no headers are included, only the payload
```

adr/ADR-52.md
```diff
@@ -0,0 +1,32 @@
...
+	// ....
+
+	// NoHeaders disable sending any headers with the body payload
+	NoHeaders bool `json:"no_hdr,omitempty"`
+}`
+```
+
+When set the server will simply send the body without any headers, headers like `Nats-Sequence` will also be unset.  
+
+This also applies to batch direct gets, where no messages in the batch will have headers. However the final zero payload message will still have the usual batch control headers.
```
MauriceVanVeen, line 32: Believe with batch get you'd still need the control headers, otherwise you couldn't detect gaps?

neilalexander (approved):
LGTM

adr/ADR-52.md
ripienaar: indeed, so then dont use this feature if you need to detect gaps :)

MauriceVanVeen: Wouldn't that mean "don't use batch if disabling headers"?
Because it could be a partial batch response otherwise?

MauriceVanVeen (approved):
LGTM

PR #379, opened 2025-09-17
Deprecate no_hdr on direct gets

ripienaar:


jpsugar-flow:
Is there a related discussion I can look at? Just curious, no objections.

ripienaar:
> Is there a related discussion I can look at? Just curious, no objections.

The implementation is not complete, we'll rather revisit later

PR #383, opened 2025-09-25
Introduce noop batch commit

ripienaar:
To support a model where users are creating a batch from a EOF based reading pattern - where they do not know if the final message is final - we introduce a way to commit the batch without storing the commit message


MauriceVanVeen (approved):
LGTM

PR #388, opened 2025-11-05
ADR-40: Custom server pool and reconnecting to specific server

piotrpio:
Signed-off-by: Piotr Piotrowski [piotr@synadia.com](mailto:piotr@synadia.com)

### DIFF
--- a/adr/ADR-40.md
+++ b/adr/ADR-40.md
@@ -129,8 +129,9 @@ Client/Connection handle.
 #### Custom server pool
 
 Client should have a way to specify custom server pool for reconnection
-attempts. When specified, client should use only those servers for reconnection
-attempts, ignoring any advertised servers from the Server.
+attempts. Client should expose an option to disable using advertised servers and
+use only the provided pool for reconnection attempts. By default, advertised
+servers are merged with the provided pool and used for reconnection attempts.
 
 New pool should be used for all reconnection attempts until the client is
 restarted or a new pool is specified. New pool is subject to the same rules as
@@ -144,26 +145,30 @@ func (nc *Conn) SetServerPool(servers []string) error
 
 #### Reconnect to specific server
 
-Client should expose an option to configure a callback that will be called
-before each reconnection attempt.
+Client should expose an option allowing users to select a specific server from
+the pool to which the client should attempt to reconnect. This can be achieved
+e.g. by a callback invoked on every reconnect attempt, allowing users to
+implement custom logic for server selection.
 
-As arguments, it should receive the list of servers that are in the pool
-(together with reconnect attempt number), as well as the most up-to-date server
-INFO.
+When choosing the server to reconnect to, the client should only consider
+servers from the pool (including advertised servers, if not disabled). Current
+server pool, together with per-server connection attempt number and the most
+up-to-date server INFO should be available to the user when choosing next
+server. User should not be able to select a server outside of the pool.
 
-The callback should return the server URL to which the client should attempt to
-reconnect (a url outside of the pool should not be allowed). Additionally, the
-callback should allow to specify whether client should include a delay before
-attempting to reconnect to the selected server (based on what's configured in
-the client options).
+Users should be able to specify whether the client should apply reconnect delay
+before attempting to reconnect to the selected server. Depending on client
+implementation, client may fall back to default reconnect delay configured on
+the connection options.
+
+Example callback-based implementation in Go:
 
 ```go
 // Return values:
-//   - url.URL: The server URL to connect to. Can be from the provided servers slice
-//     or a new server URL. New servers will be automatically added to the pool.
-//   - bool: If true, apply reconnect delay (ReconnectWait + jitter or CustomReconnectDelayCB)
-//     before attempting connection. If false, attempt connection immediately.
-type ReconnectToServerHandler func([]Server, ServerInfo) (*url.URL, bool)
+//   - url.URL: The server URL to connect to. Library should enforce that the URL is from the provided servers slice.
+//   - time.Duration: Optional reconnect delay before attempting to connect to the selected server. 
+//     If zero, reconnect will be attempted immediately.
+type ReconnectToServerHandler func([]Server, ServerInfo) (*url.URL, time.Duration)
 ```
 
 #### Detecting disconnection


adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
+
+```go
+// Return values:
+//   - url.URL: The server URL to connect to. Can be from the provided servers slice
+//     or a new server URL. New servers will be automatically added to the pool.
+//   - bool: If true, apply reconnect delay (ReconnectWait + jitter or CustomReconnectDelayCB)
+//     before attempting connection. If false, attempt connection immediately.
+type ReconnectToServerHandler func([]Server, ServerInfo) (*url.URL, bool)
+```
+
```
scottf: There also needs to be feedback to the pool.

* Give the pool discovered servers whenever the client receives them.
* Tell the pool when a connection failed or succeeded.

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
 
+#### Custom server pool
+
+Client should have a way to specify custom server pool for reconnection
+attempts. When specified, client should use only those servers for reconnection
+attempts, ignoring any advertised servers from the Server.
+
+New pool should be used for all reconnection attempts until the client is
+restarted or a new pool is specified. New pool is subject to the same rules as
+the default one - randomization unless disabled, max reconnect attempts, etc.
```
scottf: The pool should not _have to_ care about anything of the normal options or behavior, but it can care if it choses. This is the point of suppling a custom pool, to allow it to do anything it wants.

This means that randomization and the act of resolving ip addresses are part of the pool behavior.

Max reconnect attempts is not part of the pool behavior, it's part of the behavior that uses the pool. The behavior should ask for a resolved ip address, but the pool can chose what the result of a resolved ip address request is, so for instance it could give a resolved address or not when asked for an address

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
+
+Client should expose an option to configure a callback that will be called
+before each reconnection attempt.
+
+As arguments, it should receive the list of servers that are in the pool
+(together with reconnect attempt number), as well as the most up-to-date server
+INFO.
+
+The callback should return the server URL to which the client should attempt to
+reconnect (a url outside of the pool should not be allowed). Additionally, the
```
scottf: Why do we care if the reconnect to a specific server wants to connect to something outside of the pool?

What happens when a reconnect to a specific server fails?

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
 For most clients, that means having a `reconnect` method on the
 Client/Connection handle.
 
+#### Custom server pool
+
+Client should have a way to specify custom server pool for reconnection
```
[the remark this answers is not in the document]
aricart: In simpler words: calling this API replaces all the servers, with the new provided servers.

adr/ADR-40.md
aricart: ie - once the server list is set, those are the only servers - on connect, the client will learn new servers - if the client doesn't want to use gossiped servers, it should specify its option to ignore cluster updates.

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
+```
+
+#### Reconnect to specific server
+
+Client should expose an option to configure a callback that will be called
+before each reconnection attempt.
+
+As arguments, it should receive the list of servers that are in the pool
+(together with reconnect attempt number), as well as the most up-to-date server
+INFO.
```
aricart: not quite sure why we want to expose all the metrics - the callbacks for reconnecting/reconnect/disconnect have info on what is happening, and grabbing the info for their investigation is available on connection - this means that they can grab/store/use their own data to supplement the selection the "server selector".

Effectively, I would move my internal selection logic to such function and use it internally, this way how it's handled is uniform.

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
 Client/Connection handle.
 
+#### Custom server pool
+
+Client should have a way to specify custom server pool for reconnection
+attempts. When specified, client should use only those servers for reconnection
+attempts, ignoring any advertised servers from the Server.
+
+New pool should be used for all reconnection attempts until the client is
+restarted or a new pool is specified. New pool is subject to the same rules as
```
ripienaar: Need to clarify here what happens if a custom pool it set at run time and gossip learn is on, do they merge like they would normally from the initial url list or not?

adr/ADR-40.md
ripienaar: A retry like normal.

We want to clearly separate concerns here:

 * Managing the pool
 * Picking from the pool

These are different things, the callback is picking from the pool only.

adr/ADR-40.md
```diff
@@ -125,6 +126,46 @@ all reconnect options are respected.
...
+Client should expose an option to configure a callback that will be called
+before each reconnection attempt.
+
+As arguments, it should receive the list of servers that are in the pool
+(together with reconnect attempt number), as well as the most up-to-date server
+INFO.
+
+The callback should return the server URL to which the client should attempt to
+reconnect (a url outside of the pool should not be allowed). Additionally, the
+callback should allow to specify whether client should include a delay before
```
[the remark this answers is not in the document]
ripienaar: Not really true, we have various dynamic reconnect delay options already.

The use case here for example when you get information about primary and backup clusters from a service discovery system.  You would have a primary and backup cluster, you'd try the primary cluster a few times (with backoff) and finally decide on some value to switch to the standby cluster - that switch should essentially reset the backoff because now you're retrying against a new cluster.

I think probably the simple boolean here is not enough, but we do need the flexibility to influence delay+target selection as above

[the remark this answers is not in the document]
scottf: This is a different concern than the pool that is already addressed, granted in a way that the user may not be able override in certain implementations.

It may be simple enough to allow a callback for this specific behavior. For instance the Java code already has a functional callback, outside of the pool, that can have dynamic behavior, and the developer has the ability to have that callback be aware or influenced by the pool.

This is the model I propose. Keep the pool the pool. Keep other behaviors separate. It's up to the user to coordinate that behavior.

scottf:
For reference here is the current Java behavior. The pool and re/connect behavior is based on the original go reference implementation.

### Pool behavior:
* be initialized with the connection options to be able to extract information like:
    * the bootstrap servers
    * max reconnect attempts
    * whether to ignore discovered servers
    * whether to randomize
* accept discovered urls
* be able to peek at the next server - look at the next server without the pool marking it as given
* give the next server
* be asked to resolve a urls's host to a list of valid ip based utrls. Always returns a list of resolved or a list of just the original url
* be told when a connect to a server succeeded
* be told when a connect to a server failed
* be able to be asked if the pool represents the needs to upgrade to a secure connection

### Reconnect delay handler / callback:
* input is the reconnect round, starting at 1, incrementing every time the full server list is tried
* returns a length of time to wait
* default implementations leverages connection options
    * reconnect jitter for Tls/Secure connection
    * reconnect jitter for insecure connections

### Initial connect behavior:
* Peek at the next server from the pool - this gives the ability to know if you've gone through the whole list or have already tried a server. Most useful on reconnect when reconciling with max reconnect attempts.
* Get the next server from the pool
* Ask the pool to resolve the server to ip addresses.
* Try to connect to all of the resolved list
* Tell pool success of failure for the unresolved host
* Loop until all servers are tried or success;

### Reconnect behavior:
* Peek at the next server from the pool - this gives the ability to know if you've gone through the whole list or have already tried a server
* Get the next server from the pool
* Invoke the reconnect delay handler / callback and wait the expected time
* Ask the pool to resolve the server to ip addresses.
* Try to connect to all of the resolved list
* Tell pool success of failure for the unresolved host
* Loop until all servers are tried for the maximum number of reconnects or success

mtmk:
> > It feels a little too specific to me. Can we instead have a more generic callback that takes in a context and spits out a decision? The server pool selection logic can be just one implementation.
> 
> @mtmk I feel that's up to the language, as long as your client enables these behaviors (custom server pool + picking a server on reconnect). I rephrased the ADR a bit to mention a specific behaviors rather than implementation (go is there just as an example).

makes sense. thanks @piotrpio

PR #413, opened 2026-04-29
Clarify direct get to match server behaviour and remove ambiguity

ripienaar:
Also extend the mirror direct get description in ADR-59


MauriceVanVeen (approved):
LGTM

PR #415, opened 2026-04-29
Various improvements to TTL ADR after conformance

ripienaar:


adr/ADR-43.md
```diff
@@ -38,27 +39,27 @@ When a message with the `Nats-TTL` header is published to a stream with the feat
...
-This marker will also be placed for a message removed by the `Nats-TTL` timer.
+The same marker is placed when a message is removed by the `Nats-TTL` timer and that removal empties a subject. The `Nats-Marker-Reason` value is `MaxAge` in both cases — the `MaxAge` reason covers any age-based removal of the last value of a subject, whether the trigger is the stream-wide MaxAge limit or a per-message `Nats-TTL`.
 
-This behaviour is off by default unless opted in on the `SubjectDeleteMarkerTTL` Stream Configuration.
+This behavior is off by default unless opted in on the `SubjectDeleteMarkerTTL` Stream Configuration.
 
 ### Delete API Call Marker
 
 > [!IMPORTANT]
 > This feature will come either later in 2.11.x series or in 2.12.
```
MauriceVanVeen: This feature is not implemented in 2.12, nor in 2.14. It can be implemented in a future version if still desirable.

adr/ADR-43.md
```diff
@@ -67,11 +68,11 @@ Nats-TTL: 1
...
```
MauriceVanVeen: This feature is not implemented in 2.12, nor in 2.14. Opposed to the "delete marker" above, I don't think we can implement "purge marker". You can purge an entire stream or a subset, in which case: do we then try to generate N purge markers? If we limit this to "when purging all messages for 1 specific and literal subject", this could potentially be implemented in a future version if still desirable.

adr/ADR-43.md
```diff
@@ -101,10 +102,26 @@ type StreamConfig struct {
...
+ * The `Nats-TTL` header value and `SubjectDeleteMarkerTTL` setting have a minimum value of 1 second.
  * The `SubjectDeleteMarkerTTL` setting may not be set on a Mirror Stream.
  * When  `AllowMsgTTL` or `SubjectDeleteMarkerTTL` are set the Stream should require API level `1`.
  * `AllowRollup` must be `true`, stream update and create should set this unless pedantic mode is enabled.
  * `DenyPurge` must be `false`, stream update and create should set this unless pedantic mode is enabled.
- * Unless `MaxMsgsPer` equals 1 the server should treat `SubjectDeleteMarkerTTL` as the minimum for `Nats-TTL` but not reject messages that do not satisfy that. This might be changed in 2.12 depending on some internal implementation fixes in the server.
+ * Unless `MaxMsgsPer` equals 1 the server treats `SubjectDeleteMarkerTTL` as the minimum effective `Nats-TTL`. A publish with a `Nats-TTL` below this floor is **not** rejected — instead the server raises the effective TTL to the floor and rewrites the stored `Nats-TTL` header to the clamped value (formatted as integer seconds). This may change in 2.12 depending on internal implementation fixes in the server.
```
MauriceVanVeen:
```suggestion
 * Unless `MaxMsgsPer` equals 1 the server treats `SubjectDeleteMarkerTTL` as the minimum effective `Nats-TTL`. A publish with a `Nats-TTL` below this floor is **not** rejected — instead the server raises the effective TTL to the floor and rewrites the stored `Nats-TTL` header to the clamped value (formatted as integer seconds).
```

Hasn't changed in 2.12, and I don't expect it will at this point. Having both history and very different TTLs, without a minimum, will make it incredibly difficult to make look consistent for the end-user (which is the reason why we added this minimum, so the delete markers always appear consistently).

MauriceVanVeen (approved):
LGTM

adr/ADR-43.md
ripienaar: Yeah I should update to less specific terms will do

adr/ADR-43.md
ripienaar: Agree this was my feedback also but some internally insisted that yes indeed we are to publish that many markers. Seems unreasonable to me

MauriceVanVeen: Indeed, I'd agree that publishing an unbounded (potentially millions) of delete markers purely because of a full stream purge.. makes no sense to me.

But, given a 'scoped approach' of only when you're purging one specific KV key, then I still see that as viable.

(I know you agree, just expanding a bit more on my thinking)

PR #418, opened 2026-05-20
ADR: Unsafe reconfiguration of meta peer set for disaster recovery

MauriceVanVeen:


adr/ADR-61.md
```diff
@@ -0,0 +1,126 @@
...
+The subscription for this subject lives only on the system account. The `DEGRADED` token
+in the subject is intentional and signals that this endpoint is only valid when the meta
+layer can no longer make progress through normal means.
+
+The subject deliberately contains no server identifier. Every server that is a member of
+the meta group subscribes to it independently. A reconfigure request is therefore
+broadcast: each surviving server receives the request and applies the new peer set to its
+own local Raft state. The caller uses request-many semantics to gather a reply from each
+server that processed the request. There is no leader involved in this flow because the
+precondition for using this API is precisely that no leader can be elected.
```
ripienaar: What happens if some dont get it?

Personally I want the opposite, I want to do a JSZ call, get the peers and then hit each and report what happened, who got it and who didn't respond in time

adr/ADR-61.md
```diff
@@ -0,0 +1,126 @@
...
+    {"peer": "<server_name>"},
+    {"peer_id": "<server_peer_id>"}
+  ]
+}
+```
+
+`servers` is the complete new peer set, not a delta. Each entry identifies one peer using
+either:
+
+- `peer`: the human-readable server name.
```
ripienaar: I am not sure these are guaranteed unique are they?

adr/ADR-61.md
```diff
@@ -0,0 +1,126 @@
...
+- `peer_id`: the Raft peer id, which is the hash of the server name.
+
+`peer_id` is derived from `peer`. When only `peer` is supplied, the server hashes it to
+obtain the `peer_id`. When only `peer_id` is supplied, it is used as-is. When both are
+supplied, the entry is rejected unless `hash(peer) == peer_id`. The request is also
+rejected if two or more entries resolve to the same `peer_id`. Either identifier is
+accepted because, during disaster recovery, operators may have ready access to one but not
+the other depending on what tooling they're using or state they're looking at.
```
ripienaar: Ah, I guess they are - didn't know they were just hashes from the name.  For this kind of thing I prefer being strict

adr/ADR-61.md
```diff
@@ -0,0 +1,126 @@
...
+### Response
+
+The response follows the standard JetStream API envelope, with `type`:
+
+```
+io.nats.jetstream.api.v1.degraded_meta_reconfigure_response
+```
+
+The body reports the server id and name that applied the change. Errors are returned using
+the standard JS API error format.
```
ripienaar: Be good to also get back confirmation of the peer set it managed to apply I suppose

adr/ADR-61.md
```diff
@@ -0,0 +1,126 @@
...
+`peer_id` is derived from `peer`. When only `peer` is supplied, the server hashes it to
+obtain the `peer_id`. When only `peer_id` is supplied, it is used as-is. When both are
+supplied, the entry is rejected unless `hash(peer) == peer_id`. The request is also
+rejected if two or more entries resolve to the same `peer_id`. Either identifier is
+accepted because, during disaster recovery, operators may have ready access to one but not
+the other depending on what tooling they're using or state they're looking at.
+
+### Server behavior
+
+On receiving a valid request, a server:
```
ripienaar, line 104: since this is only for degraded state - it needs to verify its currently degraded? What's the risks if this happens on a non degraded set.  Like this could degrade a working cluster if you give it an invalid peer list

adr/ADR-61.md
ripienaar: We also have to be careful maybe what happens with leafs and bridged system accounts - I think those dont allow to become meta leader but do they need the full peer view and we have to handle those? We'd need a careful inventory of those and somehow be sure to hit them all.  Unless this will somehow lazily propagate once the new quorum is established?

adr/ADR-61.md
MauriceVanVeen: Indeed, this is not to be used on a properly functioning cluster, in the same way that you shouldn't call peer-remove on a server that you don't want to remove.

A couple additional checks we could add though.. that the reconfigure is only allowed if:
- There's no meta leader at the moment.
- This peer is in a state to transition to `Candidate`, i.e. allowed to start an election itself.

That would already make it safer to use.
And we should probably also revert the cluster size and quorum needed if this server gets in contact with a leader if it ended up not reconfiguring.

adr/ADR-61.md
MauriceVanVeen: Indeed, server names must be unique, since they are mapped by the server to hashes which are the peer IDs. Like `nats-0` would be `S1Nunr6R`.

adr/ADR-61.md
MauriceVanVeen: Would that be necessary?

If I send peer set X, then getting N responses back all saying they've applied set X isn't really valuable. There's no way for the peer set that's applied not to have been X, otherwise we should get an error for that server instead.

adr/ADR-61.md
MauriceVanVeen: > Personally I want the opposite, I want to do a JSZ call, get the peers and then hit each and report what happened, who got it and who didn't respond in time

This would be possible with a broadcast too though? It's proposed as a scatter-gather, so you broadcast the new peer set out to all currently connected nodes once, and you get responses from all servers whether they applied or errored. Others might time out.

We could also have a different endpoint where you can specify which server ID the request is for, but given the ['additional checks'](https://github.com/nats-io/nats-architecture-and-design/pull/418#discussion_r3273895104) I've commented about in this PR it could be that that specific server is not in a state to become candidate, and you'd need to retry on other servers then anyway. 

Seems a single broadcast is simpler? All servers that can start an election will, all online servers will respond either way, and the system will figure out who is allowed to become leader from there, and upon becoming leader it sends out the newly configured peer set.

> We also have to be careful maybe what happens with leafs and bridged system accounts - I think those dont allow to become meta leader but do they need the full peer view and we have to handle those?

I'll have to double-check, but I think it's indeed the case that leafs that extend the meta group participate in the peer set and voting, but can't become leader or start elections themselves.

I think JSZ is enough, but needs a slight addition. The `meta_cluster` field contains the `peer` field which is the `peer_id` of this server, and a `replicas` field that lists the replicas. Currently though.. the server only lists replicas if this server is the leader. I don't think that makes much sense, and we should instead always report ALL replicas even if this server is not the leader. That would allow you to request JSZ on all or a subset of servers, this should contain the leaf/bridged peers then as well.

\
\
A side-note though:
The implementation in the server probably doesn't need the peer set per se. It would only use the total peer set size for the cluster size/count and then calculate the quorum needed count from that.

I think it's good hygiene though to send the full peer set, because if we need to change the server implementation in the future, then we can do so more easily. And if the user would only send a "total cluster size" along, then we can't validate that what they sent makes sense either. And some of the checks proposed in this ADR already require the peer set to be passed to do validation prior to applying.

ripienaar: Simpler to broadcast but for this kind of tool I want to be able to, as easily as possible, be as precise as possible. So I might want to present a list of current servers and with approval go and talk to each specifically and report specifically.

This makes it clear and specific and deterministic, I know who I am calling and who will get it. vs the other isn't much easier I would still have to jsz to find the servers in order to diff the responses against known in order to report who didn't reply.

broadcast->listen is UDP.  I want TCP.

adr/ADR-61.md
ripienaar: ok

adr/ADR-61.md
ripienaar: yeah, I would also do similar pre-flight checks in the CLI but the backend needs to confirm, this is degraded use only

adr/ADR-61.md
MauriceVanVeen: Makes sense, gives more control when you want to be extra-sure the peer set is complete by querying current nodes via JSZ. JSZ can already be a PING or to specific servers, so the choice is free there.

Will adjust the proposal to target a single server only, in the same way that JSZ does by having the server ID in the subject.
Would it be worth still keeping a broadcast/PING-style endpoint, or would you say to just drop that for now?

ripienaar: We can add it back later I think if needed

sciascid:
In the ideal world, this operation would bootstrap a "new" raft group with a new id or "incarnation" number. This would prevent things like disjoint majorities and old members resurrecting and participating in the new / old cluster. I'm not sure it this is feasible though.

MauriceVanVeen:
@sciascid 
> In the ideal world, this operation would bootstrap a "new" raft group with a new id or "incarnation" number. This would prevent things like disjoint majorities and old members resurrecting and participating in the new / old cluster. I'm not sure it this is feasible though.

Yeah, I sadly think this isn't possible with how the server is implemented today. There is only a single `_meta_` group, and two distinct clusters that have had their own timeline could be joined together already today via gateways and break in various ways, with no way to guard properly.

Maybe we can look into how to improve that in the future though, because if we can store some ID on disk and replicate that.. then we should at least be able to guard against the above condition. But it'll not be easy to do for sure.

adr/ADR-61.md
MauriceVanVeen: Updated the PR to reference the additional checks I mentioned above.

adr/ADR-61.md
MauriceVanVeen: Updated to be a single per-server: `$JS.API.SERVER.DEGRADED.RAFT.META.RECONFIGURE.<server_id>`.

ripienaar: ok, somewhat verbose subject but probably ok 👍

sciascid: > A side-note though:
The implementation in the server probably doesn't need the peer set per se. It would only use the total peer set size for the cluster size/count and then calculate the quorum needed count from that.

I think the implementation should update the peer set.  As of https://github.com/nats-io/nats-server/pull/7589 the raft implementation is more careful in checking append responses, only counting messages from actual peers for quorum.  And its good to keep quorum size / peers consistent.
Also, the server should avoid accepting responses from peers that are no longer in the peer set. This would avoid cases such as: operator calls reconfigure excluding a peer, but that peer resurrects and starts replying again. Before accepting responses from that peer it needs to go through the peer addition protocol.

sciascid:
> we can look into how to improve that in the future though, because if we can store some ID on disk and replicate that.. then we should at least be able to guard against the above condition. But it'll not be easy to do for sure.

Yeah, that was my impression as well.

adr/ADR-61.md
```diff
@@ -0,0 +1,191 @@
...
+to logging the `WARN`.
+
+The advisory is published on:
+
+```
+subject: $JS.EVENT.ADVISORY.SERVER.DEGRADED.RAFT.META.RECONFIGURE
+type:    io.nats.jetstream.advisory.v1.degraded_meta_reconfigure
+```
+
+The body reports the server id and name that applied the change, its previous view of the
```
ripienaar: also need to include the client info who requested this same as JS does - if that's possible I am not sure it is as its in the same account as where it comes from? But that's the main valuable information we need.

ripienaar (approved):
LGTM just the one question about the advisory. Require @neilalexander to also approve

adr/ADR-61.md
```diff
@@ -0,0 +1,191 @@
...
+without needing to resurrect the lost peers.
+
+## Design
+
+### Endpoint
+
+A new subject is added:
+
+```
+$JS.API.SERVER.DEGRADED.RAFT.META.RECONFIGURE.<server_id>
```
sciascid: I'd consider changing the API subject to say `UNSAFE` instead of `DEGRADED`.
Just a suggestion, I don't have a good argument for this change, other than it sounds better to me :-) 
Feel free to ignore this comment.

MauriceVanVeen: I think that makes sense and clarifies that's it not normal to need to use it.
Done.

adr/ADR-61.md
MauriceVanVeen: Added

neilalexander (approved):
LGTM

adr/ADR-61.md
```diff
@@ -0,0 +1,221 @@
+# Reconfiguring the JetStream meta peer set for disaster recovery
```
ripienaar: does it really reconfigure? its more like `Unsafe meta group quorum rescue for disaster recovery`

adr/ADR-61.md
```diff
@@ -0,0 +1,221 @@
...
+working meta layer and use the normal peer-remove path to permanently drop the lost peers.
+
+## Design
+
+### Endpoint
+
+A new subject is added:
+
+```
+$JS.API.META.RESCUE
```
ripienaar, line 48: have we now recently decided that leaf nodes propagating `$JS.API.>` Is a bug we will fix? cos this will have pretty bad consequences on clusters over leaf nodes we need to prioritise that fix landing

adr/ADR-61.md
```diff
@@ -0,0 +1,221 @@
+# Reconfiguring the JetStream meta peer set for disaster recovery
+
+| Metadata | Value             |
+|----------|-------------------|
+| Date     | 2026-07-07        |
+| Author   | @MauriceVanVeen   |
+| Status   | Implemented       |
+| Tags     | server, jetstream |
```
ripienaar: Add a `2.15` tag please

adr/ADR-61.md
```diff
@@ -0,0 +1,221 @@
...
+logging the `WARN`.
+
+The advisory is published on:
+
+```
+subject: $JS.EVENT.ADVISORY.SERVER.META_RESCUE
+type:    io.nats.jetstream.advisory.v1.meta_rescue
+```
+
+The body reports the server id and name that applied the change, its previous effective
```
ripienaar, line 183: now that you've done the work maybe update here with what you are landing on for the shape

ripienaar (changes requested):

adr/ADR-61.md
MauriceVanVeen: https://github.com/nats-io/nats-server/pull/8418

adr/ADR-61.md
MauriceVanVeen: Done

adr/ADR-61.md
MauriceVanVeen: Done

adr/ADR-61.md
MauriceVanVeen: Done

ripienaar (approved):

PR #419, opened 2026-05-22
ADR-51: Document DiscardNew interaction

MauriceVanVeen:
Add a "Discard Policy Interaction" section covering how scheduling behaves on streams using `DiscardNew`:
- Stored schedules are protected from `MaxMsgs/MaxBytes` eviction.
- Single-delayed schedules always persist their delayed message and effectively replace the schedule that triggers it. This allows for patterns like `DiscardNew, MaxMsgs: 100, WorkQueue` where you have a 100 schedules that can trigger and occupy their slots, a delayed message is triggered, it is consumed and acknowledged, and then clears up a slot for another schedule.
- Repeating schedules skip to the next firing timestamp if there is no room for the scheduled message.
- Add advisories for the various events where schedules skip forward or fall back to non-source-sampled content.


ghost:
> Under `MaxBytes` the purge frees the old message's bytes, but if the new schedule is larger such that the result would exceed `MaxBytes` it is still rejected.

Oof. Not affected by this, but it is surprising behavior to lose the old message but not store the new.

MauriceVanVeen:
> Oof. Not affected by this, but it is surprising behavior to lose the old message but not store the new.

To be clear, `DiscardNew` isn't fully supported now, so this PR's purpose is to make explicit what the behavior should be.

Might need to rephrase the sentence, because the old message/schedule should NOT be purged if the new schedule is larger:
- If at limit and new schedule is equal or smaller, persist the new schedule first, then purge the old one.
- If at limit and new schedule is larger, then the publish is rejected. No purge happens.

i.e. the purge/rollup only happens after the message is persisted, if it is rejected there is no persist so also no purge/rollup.

ghost:
Oh good. That's what I expected, but the language sounded like the purge happens regardless.

adr/ADR-51.md
```diff
@@ -289,3 +291,51 @@ There are two ways to combine schedules with `Interest` retention:
...
+
+Two distinct writes are affected when the stream is full (`MaxMsgs`/`MaxBytes`, or `MaxMsgsPerSubject` under `DiscardNewPerSubject`):
+
+- **Storing the schedule.** Publishing a new schedule onto a fresh subject is a net-new message and is rejected when the stream is full, the same as any other publish. Updating an existing schedule on the same subject is a rollup (`Nats-Rollup: sub`) that replaces the prior message rather than growing the count. Under `MaxMsgs` this is net-zero on message count and so reliably succeeds even at the limit. Under `MaxBytes` the purge frees the old message's bytes, but if the new schedule is larger such that the result would exceed `MaxBytes` it is still rejected. Both `MaxMsgs` and `MaxBytes` are absolute limits that are not allowed to be exceeded.
```
jpsugar-flow:
```suggestion
- **Storing the schedule.** Publishing a new schedule onto a fresh subject is a net-new message and is rejected when the stream is full, the same as any other publish. Updating an existing schedule on the same subject is a rollup (`Nats-Rollup: sub`) that replaces the prior message rather than growing the count. Under `MaxMsgs` this is net-zero on message count and so reliably succeeds even at the limit. Under `MaxBytes` the purge frees the old message's bytes, unless the new schedule is larger such that the result would exceed `MaxBytes`, in which case it is rejected. Both `MaxMsgs` and `MaxBytes` are absolute limits that are not allowed to be exceeded.
```

adr/ADR-51.md
```diff
@@ -140,14 +141,14 @@ Here the local site would produce high frequency temperature readings into `sens
...
+| `Nats-Schedule-Target`    | The subject the message will be delivered to                                                                                                                                                                                                                                                                                                                                                                           |
+| `Nats-Schedule-Source`    | Instructs the schedule to read the last message on the given subject and publish it to the target. If no message exists on the source subject, the schedule's own body and headers is published as a fallback and an advisory is emitted with reason `source_empty` (a second fallback trigger exists under `DiscardNew`, see [Discard Policy Interaction](#discard-policy-interaction)). Wildcards are not supported. |
```
ripienaar: bit worried about the amount of advisories here, schedules can fire every second - and I think for sampled ones would be on the more frequent side - so this could result in a lot of advisories.  We already publish an empty message on such a schedule and that might be enough?

ripienaar: Advisories are for exceptional circumstances and audit trails, I dont think this situation really is either of those.  And those who get advisories are not going to care for this information its a app concern not a system concern

adr/ADR-51.md
```diff
@@ -289,3 +291,51 @@ There are two ways to combine schedules with `Interest` retention:
...
+  - **Repeating schedules (cron and `@every`)**: the schedule message must remain to keep firing, so each trigger is a net-new message. When the stream is full the generated message is rejected and dropped for that tick, exactly like any other producer hitting `DiscardNew`. The schedule message is untouched and the schedule is **not** disabled; it fires again on the next tick and succeeds once space frees up (via `MaxAge` expiry, a consumer draining the stream, or an explicit purge/delete). The server does not stall, queue, or back-fill skipped scheduled messages.
+
+For reliable schedule replacement prefer `MaxMsgs`: a replace is always net-zero on message count, whereas under `MaxBytes` a larger replacement can still be rejected.
+
+A schedule that does not produce its intended message emits an advisory so the outcome is not silent. This covers a repeating schedule that is skipped, as well as a source-backed schedule that falls back to the schedule's own content:
+
+```
+subject: $JS.EVENT.ADVISORY.STREAM.SCHEDULE.DEGRADED.<stream>
```
ripienaar: as above, dont think we need advisories here

adr/ADR-51.md
MauriceVanVeen: Removed the advisories.

MauriceVanVeen:
Superseded by: https://github.com/nats-io/nats-architecture-and-design/pull/421

PR #420, opened 2026-05-22
ADR-43: Absolute Nats-TTL & Nats-Schedule-TTL

MauriceVanVeen:
Add an absolute timestamp form to per-message TTL, just like the `Nats-Schedule: @at <ts>` pattern.

When using a schedule that needs to trigger `@every` interval, it was tricky to specify for exactly how long this should happen. You could define `Nats-TTL` as 1 week converted to seconds or a duration string, but then when the schedule stops would depend on when it was persisted. If you want it to stop on Monday before 11:00, then that's really tricky. Supporting an absolute timestamp solves that.

Similarly for use cases where you define `Nats-Schedule` to trigger on a specific `@at` timestamp and you want the resulting delayed message to expire at an exact timestamp. You would need to use the timestamp when you want it to expire, subtract the trigger time from it, convert that to a duration, and encode that in `Nats-Schedule-TTL`. Being able to pass an absolute timestamp in the per-message TTL makes these use cases significantly simpler.

There are of course guardrails:
- An absolute TTL can't be in the past.
- when used in conjunction with `Nats-Schedule: @at <ts>`, the `Nats-Schedule-TTL` (if also absolute)  needs to be at least one second after when the schedule triggers.
- Using a repeating schedule with an absolute `Nats-Schedule-TTL` is not allowed and errors, otherwise the schedule would outlive this TTL and each scheduled message after it would immediately expire.

### DIFF
--- a/adr/ADR-43.md
+++ b/adr/ADR-43.md
@@ -7,6 +7,12 @@
 | Status   | Implemented                     |
 | Tags     | jetstream, client, server, 2.11 |
 
+| Revision | Date       | Author                      | Info                                                                                       | Server Version |
+|----------|------------|-----------------------------|--------------------------------------------------------------------------------------------|----------------|
+| 1        | 2024-07-11 | @ripienaar                  | Document initial design                                                                    | 2.11.0         |
+| 2        | 2026-04-30 | @ripienaar                  | Add error codes, marker clamping behavior & 2.14 conformance notes                         | 2.14.0         |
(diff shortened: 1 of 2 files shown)


ripienaar (approved):
LGTM, need to ask the client team if they do any validation on their side of this input

MauriceVanVeen:
Closing for now, as it adds complexity (although also adds consistency), but mostly has gained no real interest thus far.

Issue #422, opened 2026-06-18
ADR-60 - Mirror/Source consumer auto created on workqueue violates non-overlapping consumer contract


roeschter:
ADR-60 explicitely mentions the non-overlapping consumer contract.  But when the consumer is auto created by a mirror or source ON a workqueue it ignores subject interest conflicts with existing consumers. The contract violation also work the other way. I can still create a local consumer when auto created one already. exists.

When two overlappong consumer exist in the workqueue message rentention seems to follow interest base semantics (which is OK) but does NOT cleanup messages when on of the consumers are delete (resulting in orphaned messages, which should not really happen on a workqueue).

Is this intentional? Its certainly not documented

roeschter:
Confirmed that this behavior is intentional. Will update ADR-60 accordingly.

PR #423, opened 2026-06-18
Workqueue auto-create consumer behavior

roeschter:
Documented the somewhat unexpected behavior of auto-created consumers on Workqueue.
Resolves - https://github.com/nats-io/nats-architecture-and-design/issues/422

### DIFF
--- a/adr/ADR-60.md
+++ b/adr/ADR-60.md
@@ -81,6 +81,8 @@ The consumer configuration will closely resemble the ephemeral push consumer var
 - Additionally, `MaxDeliver` must be set to `-1` (infinite) to ensure if some messages are lost in transit, they can
   still be reliably redelivered.
 
+
+
 ### Pre-created durable consumer
 
 An alternative to the server creating and managing ephemeral consumers for stream sourcing is the user pre-creating a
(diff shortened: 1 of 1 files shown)


adr/ADR-60.md
```diff
@@ -94,7 +96,7 @@ allowing the sourcing to temporarily stop if desired.
 
 WorkQueue streams don't allow having multiple consumers with overlapping filter subjects. This means that a durable
 consumer used for mirroring/sourcing of a WorkQueue stream, would not allow another overlapping consumer to be created
-used for a different purpose. In that case, an Interest or Limits stream should be used.
+used for a different purpose. In that case, an Interest or Limits stream should be used. Note that auto-created consumers on WorkQueue streams may violate the non-overlapping subject contract. Pre-creation fo consumers is preferred for WorkQueue.
```
jpsugar-flow, line 99:
```suggestion
used for a different purpose. In that case, an Interest or Limits stream should be used. Note that auto-created consumers on WorkQueue streams may violate the non-overlapping subject contract. Pre-creation of consumers is preferred for WorkQueue.
```

adr/ADR-60.md
```diff
@@ -112,12 +114,16 @@ for durable sourcing are met. This also allows for new use cases that the epheme
...
 - Using `ReplayPolicy=original` to allow sourcing at the speed the messages were received in originally.
 
-### Stream configuration
+### Stream configuration and consumer auto creation
 
 The stream configuration will largely remain the same, as the servers will recognize when the user tries to mirror or
 source a WorkQueue or Interest stream. However, it will be extended to include details about the consumer that should be
 used for the durable sourcing (if any).
 
+Note, that the consumer auto-created on WorkQueue stream will be able to violate the non-overlapping subject rule for WorkQueue. That multiple consumer can be created for the same subjects and messages will potentially be sourced multiple times in addition to a regular consumer on the WorkQueue.
```
jpsugar-flow, line 123: I don't understand this paragraph?

neilalexander: Sourcing consumers don't count towards the subject interest exclusivity rule, otherwise you can't have multiple sources for the same subjects and you can't combine with normal WQ consumers.

jpsugar-flow: Ah, I see. The "That multiple consumer ..." line really didn't parse.

PR #425, opened 2026-07-22
Use valid example reserved metadata

ripienaar:
When this ADR was written no reserved metadata existed yet, we do have valid examples now so use that


MauriceVanVeen (approved):
LGTM

